PR #3, opened 2024-07-20
Docker Scaffolding

falquaddoomi:
This PR adds the following:
- a Dockerfile for the frontend, so that it can be built as a container and run alongside the rest of the services
- `docker-compose.*.yml` files for various environments
- `run_stack.sh`, which facilitates building images and running the Docker Compose stack
- `backend` folder, Dockerfile, other supporting bits; includes a simple "Hello, world!" plumber API server
- placeholder folders:
    - `cluster`: SLURM Dockerfiles, config
    - `services/postgres`: configuration for the PostgreSQL instance
- updates to the `README.md` describing the above
- a template `.env` file

I'll be submitting a PR for the backend shortly that implements a more fleshed-out API and other supporting bits for the backend, but I didn't want to complicate this PR with it. I anticipate that we can get this PR approved first, then I'll submit the second backend PR that contains a more concrete skeleton of the API, and get that merged. We can then start integrating the frontend into the backend and iterate on the backend API and schema as needed.

At the moment, this builds and runs the frontend at http://localhost:5713, and the plumber API at http://localhost:9050. You should be able to access both after copying your `.env.TEMPLATE` as `.env` and invoking `./run_stack.sh`.


vincerubinetti:
Something that might be useful, since we may have many people on the team running this locally, is to open the webapp automatically when running the stack. Normally this is easy by just running `vite --open`, but it seems like it'll be more of a [pain to do that when running the vite command from docker](https://stackoverflow.com/questions/54437534/docker-open-a-url-in-the-host-browser). Perhaps you could add a simple `open localhost:5173` command near the end of `run_stack` after some timeout or perhaps waiting for some stdout output to indicate that the stack is ready.

app/package.json
```diff
@@ -3,6 +3,7 @@
   "scripts": {
     "build": "vite build",
     "dev": "vite",
+    "container-dev": "vite --host --port 5713",
```
vincerubinetti, line 6: I think we can just give these options to the regular `dev` script and avoid the duplication. I can't think of a vite flag that we'd need for the docker container but that would break things if trying to run just the frontend in isolation. Plus, once this PR is merged, I'd think doing `run_stack` would be the preferred way to run things.

README.md
```diff
@@ -0,0 +1,44 @@
+# MolEvolvR Stack
+
+This repo contains the implementation of the MolEvolvR stack, i.e.:
+- `app`: the frontend, written in Vue
```
vincerubinetti, line 4:
```suggestion
- `app`: the frontend webapp, written in React
```

README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Most of the data processing is accomplished via the `MolEvolvR` package, which
+is currently available at https://github.com/JRaviLab/molevol_scripts. The stack
+simply provides a user-friendly interface for accepting and monitoring the
+progress of jobs, and orchestrates running the jobs on SLURM. The jobs
+themselves call methods of the package at each stage of processing.
+
+## Running the Stack in Development
+
+To run the stack, you will need to have Docker and Docker Compose installed.
```
vincerubinetti, line 18:
```suggestion
To run the stack, you'll need to [install Docker and Docker Compose](https://www.docker.com/).
```

app/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+
+# use the official Bun image
+# see all versions at https://hub.docker.com/r/oven/bun/tags
+FROM oven/bun:1 AS base
+WORKDIR /app
+
+
+# -----------------------------------------------------------
+# install dependencies for dev and prod into temp directories
+# this will cache them and speed up future builds
```
vincerubinetti, line 12: Just curious, with gh actions caching, [they found that just installing from scratch every time was actually faster than caching](https://github.com/oven-sh/setup-bun/issues/14#issuecomment-1714116221). I'm guessing the cache restoring in Docker's case is much faster than in gh actions case, but just wanted to check.

vincerubinetti (approved):
This all looks good to me. Tested it out locally, and frontend, plumbr hello world, and swagger docs seem to work as expected.

Pending my other comments, feel free to merge.

app/package.json
vincerubinetti: If you do make this change, can you throw in the `--open` flag if you remember. Won't do anything when running the docker stack, but I'll probably still run the frontend directly often while developing the app and it'll help me there.

app/Dockerfile
falquaddoomi: For what it's worth, this Dockerfile, including that comment, is mostly from the official Bun docs, https://bun.sh/guides/ecosystem/docker.

I did attempt to add caching via bun's `--cache-dir` argument and a build-time volume when I saw that they hadn't actually implemented it, despite that comment. (It's a common misconception that using multi-stage builds somehow caches them outside of Docker's normal layer caching mechanism.)

Anyway, after a little testing without any form of caching I found that you're right, it's just as fast without the caching as it is with. I still find that strange, but hey, if it works I guess I can't complain.

app/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+# copy production dependencies and source code into final image
+FROM base AS release
+COPY --from=install /temp/prod/node_modules node_modules
+COPY --from=prerelease /app/main.tsx .
+COPY --from=prerelease /app/package.json .
+
+# run the app in production mode
+USER bun
+EXPOSE 3000/tcp
+ENTRYPOINT [ "bun", "run", "main.ts" ]
```
vincerubinetti, line 63: I forgot to comment on and test this. While this may work, my feeling is that we should just the built-in vite `build` and `preview` scripts, if the intention is to run the production build of the frontend. I think that will be more reliable.

app/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+# -----------------------------------------------------------
+# copy node_modules from temp directory
+# then copy all (non-ignored) project files into the image
+FROM base AS prerelease
+COPY --from=install /temp/dev/node_modules node_modules
+COPY . .
+
+# [optional] tests & build
+ENV NODE_ENV=production
+RUN bun test
```
vincerubinetti, line 36: This should be `bun run test`. `bun test` will use bun's own testing framework which is not what we want.

Also this probably isn't super necessary, because there's a github action that tests on every PR push.

app/Dockerfile
vincerubinetti: Actually, what does this intend to do. It's `main.tsx` in the copy command, and running `main.ts` (not sure where this comes from)? And it's running with bun, when this code is intended to be run in a browser. 

I guess you're already building the frontend in --prerelease and taking artifacts from there, but shouldn't it be taking the `/dist` artifacts, and hosting them on a server?

app/package.json
falquaddoomi: You're right, just adding them to `bun run dev` command in the Dockerfile seemed to work fine; I could've sworn I tried that, failed, and resorted to adding an entry in `package.json`, but I guess I must have misremembered or done it incorrectly.

About the `--open` flag, I didn't think it would work from within a Docker container, but I tried it anyway and as I suspected it didn't open a browser window. I tried a few options with similar behavior to `vite --open`, but unfortunately none were a perfect fit; I'll keep thinking about the issue, but for now I'll just use `open` on OS X, even though it always opens a new tab.

app/Dockerfile
falquaddoomi: FWIW, a lot of that Dockerfile comes from bun's example, and frankly I think it's overcomplicated. If you're ok with it I'll just replace the multi-stage build stuff with a straightforward Dockerfile, keeping your comments about using `vite` rather than `bun` in mind.

app/Dockerfile
falquaddoomi: Isn't `bun run` just running entries from `scripts` in `package.json`, which already reference `vite`? I don't see how it's any more reliable to circumvent that and invoke `vite` directly. I'm fine with changing it, but I want to understand.

app/Dockerfile
falquaddoomi: You're right, I tested and caching isn't actually making anything faster. (I thought perhaps its incredible speed was because it was secretly caching things to `~/.bun/install/cache`, which FWIW `bun` does by default, but even with that removed it's still as fast.) I went ahead and removed it.

vincerubinetti: Here's some related discussion/investigation I did a few months back. My best understanding atm is that Bun is just ridiculously fast at the dependency resolution part, and that the actual network download of packages is really not the bottleneck in package managers.

app/Dockerfile
vincerubinetti: > ENTRYPOINT [ "bun", "run", "main.ts" ]

Yes, here you're running `bun run main.ts`, right? Which is  saying run the `main.ts` script in `package.json`, which there is none as far as I see. And `main.ts` isn't a file anywhere as far as I can see either.

I'm not super clear on the Docker commands here, but I think we want something like this:

```dockerfile
COPY --from=install /temp/prod/node_modules node_modules
COPY --from=prerelease /app/dist

# run the app in production mode
USER bun
EXPOSE 4173/tcp
ENTRYPOINT [ "bun", "run", "preview" ]
```

Need to make use of the resulting `/dist` folder from `bun run build` in the "prerelease" section, and run `bun run preview` to vite serve that folder on a local server.

falquaddoomi: Sorry, my comment got mixed in with your later one. To clarify, the code from the [original bun Dockerfile](https://bun.sh/guides/ecosystem/docker) is supposed to run a "production" instance of a webserver that's hosting the frontend app. Since we're serving the app as a bundle from a reverse http proxy in production, we don't want that; I changed it to instead just build the app as a bundle when you use the '"release" target.

app/Dockerfile
vincerubinetti: Oh okay, I didn't know if that was part of aligning the frontend with some specific staging/test/production process you had decided to use for the rest of the stack.

If we want the absolute minimum, maybe we just need:

```dockerfile
FROM oven/bun:1 AS base
WORKDIR /app
RUN bun install
RUN bun run dev
```

And expose and entrypoint if needed?

We could have production "section" (not sure what you call these `from as` things in docker) where it more exactly replicates the production environment. But we'll also have the netlify PR previews to test that, and also I can count on one hand the times that there's been a difference between dev and production with vite SPAs.

app/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+    bun install --frozen-lockfile --production
+
+# -----------------------------------------------------------
+# copy node_modules from dev stage, copy entire app
+# source into the image
+FROM base AS dev
+COPY --from=install /temp/dev/node_modules node_modules
+COPY . .
+# run the app in hot-reloading development mode
+# (this requires a new script, container-dev, in package.json that sets up vite
```
vincerubinetti, line 30:
```suggestion
# set up vite
```

app/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+# to always run on port 5713)
+CMD [ "vite", "--host", "--port", "5713" ]
+
+
+# -----------------------------------------------------------
+# copy production dependencies and source code into final image
+FROM base AS release
+COPY --from=install /temp/prod/node_modules node_modules
+COPY --from=prerelease /app/main.tsx .
+COPY --from=prerelease /app/package.json .
```
vincerubinetti, line 41:
```suggestion
COPY --from=prerelease /app .
```

I think these copies need to be updated? So that vite build has access to all the source files to build them?

app/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+# to accept connections on any interface, e.g. from outside the container, and
+# to always run on port 5713)
+CMD [ "vite", "--host", "--port", "5713" ]
+
+
+# -----------------------------------------------------------
+# copy production dependencies and source code into final image
+FROM base AS release
+COPY --from=install /temp/prod/node_modules node_modules
+COPY --from=prerelease /app/main.tsx .
```
vincerubinetti, line 40:
```suggestion
```

app/Dockerfile
vincerubinetti: Nevermind, [your comment above](https://github.com/JRaviLab/molevolvr2.0/pull/3#discussion_r1697486205) clarifies that I was misunderstanding the purpose of this production "section" in the docker file (not to run a production build of the frontend, just to build a production version and make it accessible to a different part of the stack that will serve it).

app/Dockerfile
vincerubinetti: Err wait, should probably be just `COPY . .`

falquaddoomi: Duh, you're right; I'll just switch it to `COPY . .` as you suggested.

vincerubinetti (approved):

falquaddoomi:
So, I've added a very work-in-progress production configuration that actually uses the `release` target from `app/Dockerfile`. If you want to test it, you can run `./run_stack.sh prod` and then open http://localhost to see that it does in fact correctly build and serve the app bundle.

I believe I've addressed your other points, @vincerubinetti, but please let me know if I missed something. Also, thank you very much for the thorough review and commentary!

PR #4, opened 2024-08-13
Setup Playwright tests, misc fixes

vincerubinetti:
Note that the diff here is a pain because I had to do a bulk `chmod` due to a playwright setup command messing with a bunch of file permissions, and couldn't find a way to reset the permissions to their defaults (pre-playwright-command) without going through one file at a time.

- pull in various improvements from https://github.com/krishnanlab/geneplexus-app-v2/pull/37 and https://github.com/krishnanlab/geneplexus-app-v2/pull/38
- update test workflow: fix trigger, update action versions, update debugging steps, add e2e playwright testing with caching
- setup playwright for e2e and accessibility testing
- add status indicator to analysis card component
- enable msw network request mocking
- don't allow empty "lookup analysis"
- install and setup axe-core (library that powers chrome dev tools' lighthouse accessibility check)
- check all major pages of app (so far) for accessibility issues
- add basic e2e test for demonstration


vincerubinetti:
I've just added a commit that pulls in updates from https://github.com/krishnanlab/geneplexus-app-v2/pull/38, which includes unrelated fixes, but also some accessibility fixes that would affect the Axe testing here. It also adds `/testbed` to the routes that are tested with Axe.

falquaddoomi:
Hey, do you expect people to run these tests you've added locally, or are they only for the GitHub action? If I can run them locally, how do I do that? I'm getting the following when I try to do what seems obvious from reading the scripts you've defined in `package.json`:

```
[ faisala@gleipnir molevolvr2.0/app ] % bun run install-playwright
$ bunx playwright install --with-deps
[ faisala@gleipnir molevolvr2.0/app ] % bun run test:e2e                  
[WebServer] $ vite --port 1234

SyntaxError: Unexpected token 'with'

SyntaxError: Unexpected token 'with'

Error: No tests found

To open last HTML report run:

  npx playwright show-report

error: script "test:e2e" exited with code 1
[ faisala@gleipnir molevolvr2.0/app ] % bun run test    
$ bun run test:lint && bun run test:types && bun run test:e2e
$ eslint . && prettier **/*.{tsx,ts,css,html,md,json,yaml} --check --no-error-on-unmatched-pattern

Oops! Something went wrong! :(

ESLint: 9.8.0

ConfigError: Config (unnamed): Key "rules": Key "constructor-super": structuredClone is not defined
    at rethrowConfigError (/Users/faisala/Checkouts/chai/jravilab/molevolvr2.0/app/node_modules/@eslint/config-array/dist/cjs/index.cjs:303:8)
    at /Users/faisala/Checkouts/chai/jravilab/molevolvr2.0/app/node_modules/@eslint/config-array/dist/cjs/index.cjs:1098:5
    at Array.reduce (<anonymous>)
    at FlatConfigArray.getConfigWithStatus (/Users/faisala/Checkouts/chai/jravilab/molevolvr2.0/app/node_modules/@eslint/config-array/dist/cjs/index.cjs:1091:43)
    at FlatConfigArray.getConfig (/Users/faisala/Checkouts/chai/jravilab/molevolvr2.0/app/node_modules/@eslint/config-array/dist/cjs/index.cjs:1120:15)
    at /Users/faisala/Checkouts/chai/jravilab/molevolvr2.0/app/node_modules/eslint/lib/eslint/eslint-helpers.js:346:56
    at Array.reduce (<anonymous>)
    at /Users/faisala/Checkouts/chai/jravilab/molevolvr2.0/app/node_modules/eslint/lib/eslint/eslint-helpers.js:333:36
    at /Users/faisala/Checkouts/chai/jravilab/molevolvr2.0/app/node_modules/eslint/lib/eslint/eslint-helpers.js:296:32
    at Object.isAppliedFilter (/Users/faisala/Checkouts/chai/jravilab/molevolvr2.0/app/node_modules/@nodelib/fs.walk/out/readers/common.js:12:31)
[ faisala@gleipnir molevolvr2.0/app ] % 
```

vincerubinetti:
Yes you should be able to run them locally. Do you have node v22 installed?

falquaddoomi:
Aha; I wasn't using node v22 at the time, but I switched to `v22.7.0`, wiped out `node_modules`, reinstalled using `bun install`, and ran `bun run test` with the following (successful?) output:

```
% bun run test
$ bun run test:lint && bun run test:types && bun run test:e2e
$ eslint . && prettier **/*.{tsx,ts,css,html,md,json,yaml} --check --no-error-on-unmatched-pattern
Checking formatting...
All matched files use Prettier code style!
$ tsc
[WebServer] $ vite --port 1234

(node:9206) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)

Running 27 tests using 5 workers
(node:9279) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9282) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9278) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9281) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9280) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9455) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9456) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9485) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9486) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9487) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
[webkit] › home.spec.ts:4:1 › Example analyses show
(node:9557) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9556) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9579) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9584) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
  16 skipped
  11 passed (9.6s)

To open last HTML report run:

  npx playwright show-report
```

Perhaps you should specify the version of node that's required via https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines, or some other way like in the README?

vincerubinetti:
Sorry, I thought that was already in the readme. I've updated the readme to add some clarifications.

> Also, adding something to your PR description that specifies what you'd like me to test and how to do it would be nice.

What I can do from now on is make a comment near sections of the diff that are particularly complex/error-prone/high-consequence. I did this for monarch and it helped them ignore irrelevant changes. I will do this now for this PR.

app/eslint.config.mjs
```diff
@@ -0,0 +1,49 @@
+import path from "node:path";
```
vincerubinetti, line 1: New eslint config file format. Converted mostly automatically using eslint's automatic migration tool, with a few manual tweaks after. Should be equivalent to old config, though I think I added a rule or two.

app/src/App.tsx
```diff
@@ -81,11 +81,15 @@ const routes = [
         element: <Home />,
         loader: async () => {
           /** handle 404 redirect */
-          const url = window.sessionStorage.redirect as string;
-          if (url) {
-            console.debug("Redirecting to:", url);
-            window.sessionStorage.removeItem("redirect");
-            return redirect(url);
+          const path = window.sessionStorage.redirectPath || "";
+          const state = JSON.parse(window.sessionStorage.redirectState || "{}");
```
vincerubinetti, line 85: Recovers history entry `state` from 404 redirection.

app/src/App.tsx
```diff
@@ -127,6 +131,3 @@ const router = createBrowserRouter(routes, {
 
 /** query client */
 const queryClient = new QueryClient();
-
-/** prefix for localStorage keys */
-export const storageKey = "molevolvr-";
```
vincerubinetti, line 132: Not really needed anyway since storage is scoped per domain.

app/package.json
```diff
@@ -1,65 +1,70 @@
...
   "type": "module",
   "scripts": {
+    "install-playwright": "bunx playwright install --with-deps",
     "build": "vite build",
     "dev": "vite",
-    "lint": "FIX=true ./lint.sh",
     "preview": "vite preview",
-    "test": "bun run test:lint && bun run test:types",
-    "test:lint": "./lint.sh",
+    "lint": "eslint . --fix && prettier **/*.{tsx,ts,css,html,md,json,yaml} --write --no-error-on-unmatched-pattern",
```
vincerubinetti, line 8: Delete `lint.sh`, simplify, and just move contents here.

app/src/components/Header.module.css
```diff
@@ -1,20 +1,24 @@
...
+    var(--header-padding-shrink) + var(--header-logo-shrink)
+  );
+}
+
 .header {
   z-index: 2;
   position: sticky;
   top: 0;
-  padding: 20px;
+  margin-bottom: var(--header-shrink);
```
vincerubinetti, line 13: CSS var tricks to shrink header on scroll down without messing up scroll and making it jumpy. Add margin to bottom to keep overall space taken up by header the same even as it shrinks in visible size.

app/src/components/SelectMulti.tsx
```diff
@@ -50,8 +55,8 @@ const SelectMulti = <O extends Option>({
   const form = useForm();
 
   return (
-    <HUI.Listbox
-      className={classNames(classes.container, classes[layout])}
+    <Listbox
```
vincerubinetti, line 58: Upgrading headless-ui to v2. Mostly inconsequential changes.

app/src/components/Slider.module.css
```diff
@@ -67,24 +67,28 @@
...
+  top: 50%;
   left: 50%;
-  transform: translateX(-50%);
   white-space: nowrap;
   transition: opacity var(--fast);
 }
 
-.secondary-marker {
-  color: var(--gray);
+.primary-marker {
```
vincerubinetti, line 76: CSS tweaks to make various markers more readable and less overlap-y.

app/src/components/TableOfContents.tsx
```diff
@@ -30,35 +55,31 @@ const TableOfContents = () => {
...
 
-  /** active heading (first in view) */
-  const [active, setActive] = useState("");
+  /** active heading id (first in view) */
+  const [activeId, setActiveId] = useState("");
 
-  /** click off to close on small screens */
-  useClickAway(root, () => {
-    if (window.innerWidth < 1500) setOpen(false);
+  /** click off to close */
```
vincerubinetti, line 61: Tweaks to table of contents behavior. Test auto-close functionality on various screen widths. Should close when clicking off on small screens. Should close when it covers an important element on the page for more than a few moments.

app/src/main.tsx
```diff
@@ -4,19 +4,17 @@ import App from "./App";
 
 console.debug(import.meta);
 
+const mock = true;
```
vincerubinetti, line 7: Whether to enable mocking. Leave as true for now. This can eventually be toggled by something else, e.g. a URL param like we did in monarch. If doing `if (true)`, eslint/ts complains about useless conditional, so extract as var.

app/src/pages/LoadAnalysis.tsx
```diff
@@ -29,7 +29,12 @@ const LoadAnalysis = () => {
           Load Analysis
         </Heading>
 
-        <Form onSubmit={(data) => navigate(`/analysis/${data.id}`)}>
+        <Form
+          onSubmit={(data) => {
```
vincerubinetti, line 33: Don't submit if user hasn't typed in ID.

app/src/util/dom.ts
```diff
@@ -80,6 +91,43 @@ export const shrinkWrap = (element: HTMLElement | null) => {
...
-  element.style.width = width + 1 + "px";
+  element.style.width = width + "px";
   element.style.boxSizing = "content-box";
 };
+
+/**
+ * is element covering anything "important" (above anything besides a
+ * "background" element)
+ */
+export const isCovering = (
```
vincerubinetti, line 102: Util func that checks a grid of points under an element.

app/tests/accessibility.spec.ts
```diff
@@ -0,0 +1,57 @@
+import { configureAxe, getViolations, injectAxe } from "axe-playwright";
+import { expect, test } from "@playwright/test";
+import analyses from "@/fixtures/analyses.json" with { type: "json" };
+import { log } from "./util";
+
+log();
+
+/** pages to test */
+const paths = [
```
vincerubinetti, line 9: Accessibility checks. Simply (for now) goes through a list of routes, and checks the whole page for accessibility issues. Essentially like if you were to manually run lighthouse in chrome dev tools on each page.

app/src/main.tsx
falquaddoomi: Nice; I actually like having this extracted as a variable so it can be quickly toggled, versus looking for an `if (true)` statement somewhere in the code. I also like the idea of having it being toggled without having to change the code; perhaps it could start as an env var and also get its value from the URL later, as you suggested.

falquaddoomi:
> What I can do from now on is make a comment near sections of the diff that are particularly complex/error-prone/high-consequence. I did this for monarch and it helped them ignore irrelevant changes. I will do this now for this PR.

Thanks; I realize this took effort, and it is helpful. What I meant, though, was just some simple directions for things you'd want tried out by a reviewer, e.g. "Run `bun run test` and see that the tests complete" or "try out feature X in widget Y". I'm not much of a frontend dev anymore, and just simple things like "use node version X" or "run this command to launch the test suite" would save me a little time.

You're of course welcome to continue to describe the significant code changes in your PR and why you made them, but I don't expect you to.

vincerubinetti:
Gotcha. I'll still try to do the comment-by-comment when possible, still seems helpful on large PRs so you can sort of skip over lower risk/impact things. And if not I'll have a paragraph or so summarizing the salient things.

app/src/util/dom.ts
falquaddoomi: Interesting solution, and it seems to work well.

I'm not suggesting changing it, but if you're interested you might consider trying the [intersection observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API), which seems to have good support from all the major browsers. Basically it allows you to register a target element that gets checked against all the elements under some node in the DOM (by default it's the browser window, i.e. all elements) for intersections and fires a callback when an intersection occurs. Supposedly it's more performant than checking yourself, but since you're just sampling a few points I think your solution is fine.

falquaddoomi (approved):

app/src/util/dom.ts
vincerubinetti: I'm not sure if ResizeObserver can be used in this manner, because although you can specify a root element other than the viewport, it seems that the root must be an ancestor. That is, the element being covered (let's say an svg chart element) would have to be a child of the table of contents. 

Idk if there is a browser API I could use to make this less kludgy. I could do a simple bbox overlap check, but then I'd have to do it against all the elements in the viewport. It might end up being faster than this method though.

falquaddoomi:
Just FYI, I've looked through the changes and I still approve it all. Feel free to merge.

app/src/util/dom.ts
vincerubinetti: I switched to the bbox method. I found the elementsFromPoint method to range from < 1ms to 10ms, and the bbox method to always be < 1ms. Since this function is run in a very debounced fashion, this should never really matter, but it could matter if we needed to call it on every scroll event in the future. The bbox method, with the default "important" element selector, should definitely do less work on average than elementsFromPoint.

Bbox method, at least with the default `section > *` selector, will be less "accurate", in the sense that it's based on DOM hierarchy and not actual page layout, which are not always the same. For example:

```html
<section>
  <!-- narrow box in center of screen. gets checked by util func by default, b/c is a direct descendant of section -->
  <div style="width: 100px">
    <!-- image that is "popped out" of regular document flow, positioned outside its parent div, floating on the left side of the screen. not checked by the util func by default, but could still be under the floating table of contents panel -->
    <img style="position: fixed; left: 0"/>
  </div>
</section>
```

This is unlikely though. If it becomes a problem, could change the selector to `section *` (would have to check many more elements, becoming possibly slower than the elementsFromPoint method).

PR #5, opened 2024-09-04
Adds draft backend API, db integration, db schema migration

falquaddoomi:
This PR replaces the "dummy" API from the previous PR with one that does the following:
- interacts with the postgres server to store and retrieve analysis objects, served from the `analysis` endpoint
- implements other very WIP endpoints, e.g. `stats`, just for testing
- adds database schema migrations via [Atlas](https://atlasgo.io/)
- adds more verbose JSON serialization *(to get around Plumber just crashing when it encounters unserializable values)*

Other non-API-related changes:
- adds WIP test harness that allows tests to be located in subdirectories
- adds `shell` control command to `run_stack.sh` that starts a shell in the backend container
 
Things you might want to test:
- ensure that everything you need to create your `.env` file is in `.env.TEMPLATE`
- ensure that you can start the stack by running `./run_stack.sh`
- ensure you can see the API at http://localhost:9050
- try making a POST request to analyses, possibly via the Swagger UI at http://localhost:9050/__docs__/
- after the stack is up, try running `./run_stack.sh shell` in another window to get a shell into the backend
  - try running the test harness script from there, e.g. `./run_tests.sh`


vincerubinetti:
Looking at this now.

Do you know if Plumbr can export OpenAPI types? If so, I could [automatically generate typescript types](https://www.npmjs.com/package/openapi-typescript) for everything. I found [this page](https://rdrr.io/cran/plumber/src/R/openapi-types.R) but I can't tell if it's Plumbr source code, a tutorial/example, a Plumblr plugin, or something else completely.

Also just curious about the docs endpoint being `/__docs__`. Is that a Plumbr requirement or something? Is there a naming conflict we're worried about here?

vincerubinetti:
Okay, looks like you could call `getApiSpec`. Not sure if this returns a path or the actual json of the spec (docs aren't clear about that...), but you could write that to e.g. `/backend/api/spec.json` (my recommendation). The backend would still need to run to generate this, but now that I think about it there's probably no way around running it.

backend/README.md
```diff
@@ -0,0 +1,29 @@
+# MolEvolvR Backend
+
+The backend is implemented as a RESTful API over the following entities:
+
+- `User`: Represents a user of the system. At the moment logins aren't
+required, so all regular users are the special "Anonymous" user. Admins
+have individual accounts.
+- `Analysis`: Represents an analysis submitted by a user. Each analysis has a unique ID
+and is associated with a user. analyses contain the following sub-entities:
```
vincerubinetti, line 9: Very persnickety comment, but perhaps all of these sub-entities should be have the `Analysis`- prefix. Or in cases where "Analysis" is already "in the namespace", you could omit it, e.g. in types:

```ts
type Analysis = {
  submission: Submission; // i guess what i'm calling "inputs" in the frontend so far
  status: Status;
  result: Result;
}
```

I'm going to try to keep type, variable, and in-UI names consistent with what you use on the backend, so this will trickle down.

backend/README.md
```diff
@@ -0,0 +1,29 @@
...
+    - `Submission`: Represents the submission of a Analysis, e.g. the data
```
vincerubinetti, line 10: Suggestion:

> `Submission`: Represents the submission of a Analysis, i.e. the input data and parameter selections by the user and the system.

backend/README.md
```diff
@@ -0,0 +1,29 @@
...
+       itself as well the submission's parameters (both selected by the 
+       user and supplied by the system).
+    - `AnalysisStatus`: Represents the status of a Analysis. Each Analysis has a status
+    associated with it, which is updated as the Analysis proceeds through its
+    processing stages.
+    - `AnalysisResult`: Represents the result of a Analysis.
```
vincerubinetti, line 16:
```suggestion
    - `AnalysisResult`: Represents the result of a Analysis, i.e. the output data of interest and other metadata like time it took analysis to run.
```

backend/README.md
falquaddoomi: Erm, sorry about that README; I've changed nearly all the content there in the upcoming PR, and the part you've commented on isn't there anymore. Perhaps let's just design the thing first and then write the documentation for it once things are more stable.

vincerubinetti:
Other than my above comments, this seems to work as expected, though I have no experience with Plumber. Maybe Dave or David could look over this?

vincerubinetti (approved):

falquaddoomi:
> It works for me. Very cool. One last thought, maybe we rename /app to /frontend, and then use "app" to refer to the whole stack. I saw /app in the Plumber source code, and I was a little confused why it was referring to the frontend, but now I figure that has something to do with it all being mounted in /app when you shell into the container?

Right, I can see how that's confusing...kind of out of habit, I put application-specific code in a container in a folder in the container's filesystem called `/app`; it's not referencing the frontend.

I do like the idea of changing the `app` folder in the repo to `frontend`, to clarify that it's the frontend and not the whole stack.

falquaddoomi:
> Okay, looks like you could call getApiSpec. Not sure if this returns a path or the actual json of the spec (docs aren't clear about that...), but you could write that to e.g. /backend/api/spec.json (my recommendation). The backend would still need to run to generate this, but now that I think about it there's probably no way around running it.

Right, the OpenAPI spec is generated via reflection at runtime, which means the backend has to be running. I'll pull the OpenAPI spec as a file and add it to this PR.

I will say that the OpenAPI spec Plumber generates by default has limited usefulness; R is famously loose with its typing, so Plumber can't extract much of a schema for the return value from the API, and it can barely do it for the parameters, too, since base R doesn't support type hinting. Plumber's alternative is for you to add an annotation to each endpoint, `#* @response <name> <openapi-formatted response>` that it then basically dumps back into the OpenAPI spec.

Here are two possible options:
1. We manually create the OpenAPI spec as JSON/YAML, instruct Plumber to use it via [set_api_spec()](https://www.rplumber.io/reference/pr_set_api_spec.html), and ensure the actual API conforms to it by calling [validate_api_spec()](https://www.rplumber.io/reference/validate_api_spec.html).
2. I add the `#* @response` annotations to each endpoint as Plumber describes in its [block annotations docs](https://www.rplumber.io/articles/annotations.html#block-annotations), we pull the resulting OpenAPI spec file, and put it somewhere in the filesystem for the frontend build tools to find.

Option 1 is more straightforward IMO and makes the OpenAPI spec the source of truth rather than an artifact. While option 2 makes the code more verbose and requires a step where the spec is extracted into a file, there's something to be said for the response documentation being near the code that generates it.

vincerubinetti:
A third option is just to not deal with a spec at all, and I'll just manually make my typescript as I was planning to do anyway. 

Do you think it would
help you at all on the backend to have a spec? If you get little or no value from that, I'd vote for option 3.

vincerubinetti:
Let me try converting the dump you just pushed to typescript and seeing how useful it is as well.

vincerubinetti:
> Let me try converting the dump you just pushed to typescript and seeing how useful it is as well.

Okay I think it may be of some help to me, even if it's just for getting the names of the endpoints and being able to get a [TS API client](https://openapi-ts.dev/openapi-fetch/#basic-usage) out of it. I'm thinking I'll probably use it, even if I have to supplement it with hand-written types.

I get the benefits of option 1, but I think writing an OpenAPI spec by hand for all of the endpoints and complex data structure's we're going to have is going to become a burden. It's fairly verbose and repetitive. Option 2 seems like it's a Plumber special syntax that's a little more terse? And I'm assuming since it gets converted to a spec anyway, it has the same level of checking (R type hinting? compile warnings?) as option 1, if any at all.

I suppose ultimately, someone is going to have to do the work of defining the structure of this data, it's just a question of who. I'm happy to take on that task (especially since the frontend is enforcing this requirement by using TS), especially if doing the type annotations at the backend level doesn't really benefit you in any way. Would it help to have those annotations collocated with those endpoint functions? I feel like most of the structure info is really going to be coming from the package level anyway. Having the Swagger UI be more descriptive and type-safe would be a nice benefit though

Maybe I just try out manually converting the spec to TS for a while -- seeing how useful and accurate it is -- before we make the effort to do one of the options and/or automate the TS conversion.

In any case, let's hold off for this PR at least, this is getting too in the weeds. I might copy this discussion to an issue. 

---

Regarding "app" -> "frontend", since you said you have some upcoming changes relevant to this stuff, maybe we can wait until that's done before renaming the folder and associated docker commands. Possibly in a new PR.

falquaddoomi:
> A third option is just to not deal with a spec at all, and I'll just manually make my typescript as I was planning to do anyway.

The third option sounds good, at least for now. Maybe once things have crystallized a bit more we can go back to looking at a server-side spec, if only for documenting the API and for validation, not necessarily for your frontend.

PR #14, opened 2024-09-28
Network visualization

### LINKED ISSUE #6: Network viz
Must-haves:

- [ ] customizable node labels, sizes, colors
- [ ] customizable edge labels, sizes, colors, directionalities
- [ ] selectable nodes
- [ ] selectable edges
- [ ] legend with key and details about selected node/edges
- [ ] self-referencing node (edge "loop")
- [ ] multiple edges between same nodes  ("stacked edges")
- [ ] download as high-res png, csv, cytoscape, etc.
- [ ] expand/collapse toggle
- [ ] "max nodes" slider (sorted by some metric)
- [ ] "max edges" slider (sorted by some metric)

Nice-to-haves:

- [ ] snap to grid
- [ ] more physics control, e.g. pin/unpin all nodes
- [ ] settable width/height
- ... other features from https://het.io/search/?source=36083&target=28137&metapaths=CdGaDrD

vincerubinetti:
Closes #6 (except for "nice to haves", which we can revisit later)

- fix playwright install action
- add some nice options to flex component
- network viz
  - settle on cytoscape library
  - customizable colors/sizes/etc
  - legend and key
  - selectable nodes/edges
  - max nodes slider
  - expand button
  - download options
  - add all available built-in and plugin layout algorithms, except ones with only a few dozen stars on github and ones that crash. explicitly set some key options for each algo to make them look sensible.
- add network viz to top of testbed page, with fake generated data
- util func to pick key colors
- download util funcs


frontend/src/pages/Testbed.tsx
```diff
@@ -61,14 +63,81 @@ const logChange = (...args: unknown[]) => {
...
+      id: uniqueId(),
+      label: sample([
+        "Lbl.",
+        "Label",
+        "Long Label",
+        "Really Long Label",
+        undefined,
+      ]),
+      type: sample([
+        "gene",
```
jananiravi, line 79: @vincerubinetti at what stage do these attributes need to be defined? e.g., circles for genes, rounded-edge-rectangles for diseases, triangles for drugs, etc?

Also, right now we only have protein domains. This may change down the line, though (or when adapted to other projects).

frontend/src/pages/Testbed.tsx
```diff
@@ -61,14 +63,81 @@ const logChange = (...args: unknown[]) => {
...
+        "disease",
+        "compound",
+        "anatomy",
+        "phenotype",
+        undefined,
+      ]),
+      strength: sample([0, 0.1, 0.02, 0.003, 0.0004, 0.00005, undefined]),
+      extra: sample(["cat", "dog", "bird"]),
```
jananiravi, line 87: do you need example networks | or can you pull this from one of the example molevolvr runs?

frontend/src/pages/Testbed.tsx
```diff
@@ -61,14 +63,81 @@ const logChange = (...args: unknown[]) => {
...
+        "Lbl.",
+        "Label",
+        "Long Label",
+        "Really Long Label",
+        undefined,
+      ]),
+      source: sample(ids)!,
+      target: sample(ids)!,
+      type: sample([
+        "causes",
```
jananiravi, line 105: Similarly, when should these be defined? Accordingly 'define attributes' can be an open issue that reminds us to get back to these later.

frontend/src/util/download.ts
```diff
@@ -26,17 +39,42 @@ export const download = (
...
-  download(stringify(data), filename, "text/csv;charset=utf-8", "csv");
+  download(
+    stringify(data, { header: true }),
+    filename,
+    "text/csv;charset=utf-8",
+    "csv",
+  );
+
+/** download table data as tsv */
+export const downloadTsv = (data: unknown[], filename: string | string[]) =>
```
jananiravi, line 50: currently, our network script only carries three columns
- node1, node2, direction
This info is used to calculate node freq, edge freq, auto-loops, etc. currently, there's no additional metadata. If a particular input format is required for this cytoscape network, then we need to make sure the molevolvr outputs return the required values.

jananiravi (approved):
High-level quick look -- seems OK. I've asked a few clarifying Qs to prepare us for how this will be used within MolEvolvR.

frontend/src/pages/Testbed.tsx
vincerubinetti: The fake data I'm generating here should explicitly stay fake, because I've set it up to test various cases (e.g. wide range of values, looped nodes, stacked edges, etc.)..

This is just for the testbed though. I will eventually ask for sample network data from a real analysis, and I'll use that as "mock" data that will be used when running tests or available behind a flag (e.g. put `?mock=true` in the url to run the frontend with mock data and not hit the real API). At that point, we'll be able to see that real data on the Analysis results page, and we'll also be able to make visual tweaks to suit the actual data.

frontend/src/pages/Testbed.tsx
vincerubinetti: What stage as in what stage of our development process? Or stage as in, flow of data through the frontend?

Right now, I've designed this as a generic component. That is, not too hard-coded to the exact specific use-case you might have in mind right now. This could allow reuse in a different project, or in this project if you ever want to use the network viz for some different data.

frontend/src/pages/Testbed.tsx
vincerubinetti: I think you're asking specifically about the `type` field here?

This is just an arbitrary category that the network component uses to assign different colors to different nodes/edges. It doesn't care what the type is, or have any context for what it means; I've tried to separate the concerns there. 

I was assuming there'd be some use in this, e.g. assigning different species. It could also just as easily be all one color, or random colors. Also if for some reason in the future you wanted to make sure a certain that a specific color was mapped to a specific type, we could easily change it to accept a color mapping instead of the component doing it automatically.

Related, in an above comment you mentioned having different shapes as well. Cytoscape supports that, so we could have a second "type". However I was hoping to reserve different shapes for going along with the colors as a color-blind-accessible option, like so:

image

frontend/src/util/download.ts
vincerubinetti: I wanted to make sure arbitrary data download was supported by these util funcs.

There shouldn't be any bending-over-backwards to fit Cytoscape. The way I try to write these components is to take a logical, general purpose set of inputs (e.g. "type" rather than "species", or "strength" rather than "p-value"), and translate that to whatever the library we're using needs. So later if we want to switch off of Cytoscape to something else, we need only change this one component.

I think this warrants some discussion in our next meeting, and will also become clearer as we start to hook up real endpoints.

vincerubinetti:
I'll add shapes (tied to colors which are mapped to unique `type`s, for color-blindness) in a future PR.

Also more filters can/will be added when this network component starts to be used in-situ. E.g., only show nodes with p-value between A and B. These filters will exist outside the component, since they are more specific and not generic to all use-cases of the component. That is, the sliders will be outside of the component, and the node/edge data will be filtered appropriately before being passed to the component.

PR #15, opened 2024-10-01
Dark mode

vincerubinetti:
This is based off of #14 . I'll rebase it once that PR is merged so only https://github.com/JRaviLab/molevolvr2.0/pull/15/commits/32e2a446bb211d9f0cc7ec1ef3bf2cdaf10b3b66 remains.

- use first-party `@axe-core/playwright` instead of `axe-playwright`
- add type packages for network viz layout algorithms
- add dark mode toggle button and sync with local storage
- add dark mode theme colors
- replace a few hard-coded colors in a few spots with css theme vars
- move testbed network randomized data generation outside component so it only changes on page refresh, not every component re-render
- change "get css var" util funcs to be a reactive hook, which will update on dark mode toggle


frontend/src/components/Network.tsx
```diff
@@ -184,7 +200,7 @@ const layouts = [
     label: "Random",
     padding,
     boundingBox,
-  },
+  } satisfies RandomLayoutOptions & LayoutExtras,
```
falquaddoomi, line 203: This is the first time I've seen `satisfies`; I read up on it and it seems that it validates the type of the preceding expression matches the type you give, as opposed to narrowing the type by explicitly assigning it.

This all looks fine, but just for my understanding/curiosity: I'm guessing you're using it here so that you can validate the types of individual entities of this array? Since each element appears to satisfy a different type, I suppose you can't just assign `layouts` a type, right?

falquaddoomi (approved):
Neat! Seems pretty straightforward. I tried it out in the Netlify preview and it seemed to work as expected, too.

frontend/src/components/Network.tsx
vincerubinetti: That's right, it checks that it conforms without casting it. Casting it would make it that type where it's being consumed, but wouldn't check that it actually conforms to that type where it's being set. There is a `BaseLayoutOptions` (or something like that) type that other layout types are supposed to inherit from, and I could set e.g. `const layouts: BaseLayoutOptions[]` (ensures that type when setting and consuming), but I found some incompatibilities between it and the other type-defs. This is all very old and rough; most of these libraries were written when TS was still very new. It also becomes a hairy union of `(BaseLayoutOptions & (LayoutAOptions | LayoutBOptions | ... ))[]` that ends up obfuscating instead of clarifying.

So, this is kind of the only option, given how the types are constructed, across separate libraries.

Issue #18, opened 2024-10-03
Limit pan in network visualization


vincerubinetti:
There is a limit to how far you can zoom in/out in the network component, but no limits for how far you can drag the view away from the content.

Add pan limits such that the user can't drag the view too far away from nodes/edges. This is not a built in feature in Cytoscape.js, the library we're using for the visualization:

https://github.com/cytoscape/cytoscape.js/issues/1329

This will likely involve listening for the [`viewport` event](https://js.cytoscape.org/#events/graph-events) and limiting the pan accordingly.

Reference:
https://stackoverflow.com/questions/55057075/how-to-check-if-my-cytoscape-js-graph-is-out-of-view-even-partially-after-pann

daud99:
@vincerubinetti can you assign this one to me?

daud99:
This one looks exciting!

daud99:
@vincerubinetti

PR #28, opened 2024-10-06
Limit pan in network visualization (#18)

daud99:
Fixing the #18

Problem: Users could pan graphs infinitely in any direction, sometimes losing the graph completely from view. We needed different handling for small graphs (that fit viewport) vs large graphs (that exceed viewport).

Solution:
1. For small graphs that fit viewport:
   - Keep graph fully visible at all times
   - Snap back with animation if user tries to pan outside bounds
   - Center graph if no valid position is found

2. For large graphs that exceed viewport:
   - Calculate graph's center and dimensions relative to viewport
   - Allow panning up to 10px beyond graph edges in any direction
   - Use center-based math to ensure uniform panning limits on all sides
   - Add smooth animations when bringing graph back into valid position
   - Fallback to centering if graph somehow gets completely hidden

The solution ensures users never lose track of their graphs while still maintaining fluid pan interactions. The uniform 10px edge padding gives just enough peek space without letting users pan too far.

### DIFF
--- a/frontend/src/components/Network.tsx
+++ b/frontend/src/components/Network.tsx
@@ -62,6 +62,11 @@ import { lerp } from "@/util/math";
 import { formatNumber } from "@/util/string";
 import classes from "./Network.module.css";
 
+type Pan = {
+  x: number;
+  y: number;
+}
+
 type Node = {
   /** unique id */
   id: string;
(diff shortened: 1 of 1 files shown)


daud99:
@jananiravi @vincerubinetti 
 
I found it a bit tricky since there wasn't a straightforward solution available online. Most of the related issues are either still open or not pointing in the right direction.

I believe the solution I developed successfully places limits on the panning that users can do. I've tested it with various types of network, and it’s working well for me.

Could you kindly test it and share your feedback? I would love to know your thoughts.

Regards

vincerubinetti:
Thank you for your effort on this. Here are my thoughts.

 - I didn't explicitly say this upfront, but the value/impact of this issue is low enough that it wouldn't warrant writing and maintaining 100+ lines of code.
- I found some oddities in the behavior. It can get stuck beyond the bounds sometimes. Also if someone wants to zoom into a particular node at the edge of the content bbox, it's a little bit harder. 
- This is maybe a subjective thing, but for such a continuous thing like this, I'd prefer to not have an animation. It felt weird/sluggish to be constantly trying to animate back to center when dragging.

See what I replaced the behavior with. I think this is simpler and has better UX. The code handles just one case, more consistently, and it limits the pan immediately and gives the user more freedom.

There are a couple other small stylistic things in your original solution that could be improved which I'll also comment on briefly, in your first commit.

frontend/src/components/Network.tsx
```diff
@@ -62,6 +62,11 @@ import { lerp } from "@/util/math";
 import { formatNumber } from "@/util/string";
 import classes from "./Network.module.css";
 
+type Pan = {
```
vincerubinetti: A small thing, but you could extract/derive this prop instead of redefining it. See below.

frontend/src/components/Network.tsx
```diff
@@ -423,6 +428,147 @@ const Network = ({ nodes: _nodes, edges: _edges }: Props) => {
...
+      const height = cy.height();
+      const zoom = cy.zoom();
+      const viewport = cy.extent();
+      const boundingBox = cy.elements().boundingBox();
+      const pan = cy.pan();
+  
+      /* Check if the graph can fit inside viewport */
+      if (viewport.h > boundingBox.h && viewport.w > boundingBox.w) {
+        const transformedBounds = {
+          x1: boundingBox.x1 * zoom + pan.x,
```
vincerubinetti: Some duplication here with code below.

frontend/src/components/Network.tsx
```diff
@@ -423,6 +428,147 @@ const Network = ({ nodes: _nodes, edges: _edges }: Props) => {
     [_edges, nodes, minEdgeStrength, maxEdgeStrength, edgeColors],
   );
 
+  const setupCytoscapePanLimits = () => {
+    let panning = false;
+    let currentPan: Pan | undefined = undefined;
```
vincerubinetti: `Pan` here could be something like `ReturnType<typeof cy.pan>`, or the Cytoscape types might just export a `Pan` type that you can just import (they probably/hopefully do).

frontend/src/components/Network.tsx
```diff
@@ -447,6 +593,7 @@ const Network = ({ nodes: _nodes, edges: _edges }: Props) => {
           .map((element) => element.data() as Node | Edge),
       ),
     );
+    setupCytoscapePanLimits();
```
vincerubinetti: Very minor stylistic comment, but keeping with the naming style of the rest of the code base so far, this could just be named `setupPanLimits`, or even just `limitPan`.

daud99:
@vincerubinetti Thankyou so much for taking out your time and reviewing it. You're right. You have really made it quite minimal and optimal. Thanks for sharing it.

:) 

Regards

vincerubinetti:
Thanks for your work here. You were on the right track, and as you said, it was not very straightforward. Really this should be something that is built into Cytoscape.

PR #31, opened 2024-10-10
Abandon apca check

### LINKED ISSUE #22: Implement color contrast accessibility test
`accessibility.spec.ts` currently has the following:

```
/** NOT WORKING YET https://github.com/StackExchange/apca-check/issues/143 */
registerAPCACheck("bronze");

// ...

      const { violations } = await new AxeBuilder({ page })
        /** https://github.com/dequelabs/axe-core/issues/3325 */
        .options({ rules: { "color-contrast": { enabled: false } } })
        .analyze();
```

See the two github issues linked there for full details. In summary:

We have disabled the standard "color-contrast" accessibility test, because it is frankly not very accurate or good. I want to replace it with the newer and better [advanced perceptual contrast algorithm](https://www.google.com/search?q=apca+%28advanced+perceptual+contrast+algorithm), but I'm having trouble integrating that `StackExchange/apca-check` library with Playwright.

This seems to be challenging to fix; see the first issue for problems I've encountered and what I've tried.

vincerubinetti:
Closes #22 as wont-fix. See https://github.com/StackExchange/apca-check/issues/143

- update expand/collapse icons
- remove `apca-check` as it wasn't working anyway
- re-enable normal/current WCAG v2 (or whatever) `color-contrast` rule (which is very inaccurate and not very useful), and make it a warning instead of a failure

See this issue, which _might_ cover APCA:
https://github.com/JRaviLab/molevolvr2.0/issues/33


falquaddoomi (approved):
The PR looks reasonable. Too bad about apca-check not working and the general lack of good contrast testing in the other accessibility suites, but as you said Lighthouse could be an alternative.

PR #38, opened 2024-10-17
Extract cytoscape shapes

### LINKED ISSUE #20: Add different node shapes to network visualization
The network visualization currently indicates different "types" of nodes/edges by using different colors. For accessibility, we'd like to have a different shape for each unique color. Basically, we want to map each [unique color here](https://github.com/JRaviLab/molevolvr2.0/blob/main/frontend/src/util/color.ts) to a different [cytoscape node shape](https://js.cytoscape.org/#style/node-body), then use those shapes for the nodes based on their `type` in the same way that color is done.

We unfortunately can't really do this with the edges. The only non-color styling we could apply to lines to distinguish different types is dash patterns, but I think doing that will actually make the visualization harder to read/understand.

A challenge for this is that the legend will need to show these shapes. Currently it's just showing simple circles. try to find a way to extra the raw point geometry of all the shapes from the Cytoscape library, so we don't have to re-define them all ourselves. The source code where Cytoscape generates these seems to be here:

https://github.com/cytoscape/cytoscape.js/blob/bce954c2bc34b88c46fe4503af9eb8b824fb6ffd/src/extensions/renderer/base/node-shapes.js#L522

Perhaps we can import that `BRp` object somehow.

AleruDivine:
Closes #20

- Extract node shape raw geometric points from Cytoscape's base renderer
- Create and export `nodeShapes` object with all predefined shapes (filtering out the `makePolygon` function)

### DIFF
--- a/frontend/src/util/shapes.ts
+++ b/frontend/src/util/shapes.ts
@@ -1,23 +1,19 @@
 import cytoscape from "cytoscape";
+import { omit } from "lodash";
 
-export type NodeShape = {
-  name: string;
-  points?: number[];
-};
+type NodeShapesRecord = Record<cytoscape.Css.NodeShape, { points?: number[] }>;
 
 type CytoscapeRenderer = {
   prototype: {
     registerNodeShapes: () => void;
-    nodeShapes: Record<string, NodeShape>;
+    nodeShapes: NodeShapesRecord;
   };
 };
 
 /** initialise cytoscape base renderer extension and access only nodeShapes */
-export const nodeShapes = ((): Record<string, NodeShape> => {
+export const nodeShapes = ((): NodeShapesRecord => {
   const baseRenderer = cytoscape("renderer", "base") as CytoscapeRenderer;
   baseRenderer.prototype.registerNodeShapes();
   const shapes = baseRenderer.prototype.nodeShapes;
-  return Object.fromEntries(
-    Object.entries(shapes).filter(([key]) => key !== "makePolygon"),
-  );
+  return omit(shapes, "makePolygon");
 })();
--- a/frontend/src/util/shapes.ts
+++ b/frontend/src/util/shapes.ts
@@ -1,0 +1,23 @@
+import cytoscape from "cytoscape";
+
+export type NodeShape = {
+  name: string;
+  points?: number[];
+};
+
+type CytoscapeRenderer = {
+  prototype: {
+    registerNodeShapes: () => void;
+    nodeShapes: Record<string, NodeShape>;
+  };
+};
+
+/** initialise cytoscape base renderer extension and access only nodeShapes */
+export const nodeShapes = ((): Record<string, NodeShape> => {
+  const baseRenderer = cytoscape("renderer", "base") as CytoscapeRenderer;
+  baseRenderer.prototype.registerNodeShapes();
+  const shapes = baseRenderer.prototype.nodeShapes;
+  return Object.fromEntries(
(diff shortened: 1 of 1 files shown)


frontend/src/util/shapes.ts
```diff
@@ -0,0 +1,23 @@
...
+    nodeShapes: Record<string, NodeShape>;
+  };
+};
+
+/** initialise cytoscape base renderer extension and access only nodeShapes */
+export const nodeShapes = ((): Record<string, NodeShape> => {
+  const baseRenderer = cytoscape("renderer", "base") as CytoscapeRenderer;
+  baseRenderer.prototype.registerNodeShapes();
+  const shapes = baseRenderer.prototype.nodeShapes;
+  return Object.fromEntries(
```
vincerubinetti, line 20: We have lodash installed, so you could just use `omit(shapes, "makePolygon")` here.

frontend/src/util/shapes.ts
```diff
@@ -0,0 +1,23 @@
+import cytoscape from "cytoscape";
+
+export type NodeShape = {
+  name: string;
+  points?: number[];
```
vincerubinetti, line 5: Are you sure these type defs can't be extracted from Cytoscape directly, instead of redefining them?

iamTruthy: After checking the cytoscape type definition, I didn't see any types matching. The type definition returned when cytoscape is initialized with two string arguements is `unknown`. Also, in the cytoscape's type defintion file, there exists a `nodeShapes` type but it doesn't match the returned type from:
```ts
const shapes = baseRenderer.prototype.nodeShapes;
```
But please I'm fully open to suggestions.Thank you.

AleruDivine: Thank you for the question. 

The current type definitions we're using are partially from Cytoscape and partially custom because:

1. For the shape names, we're using Cytoscape's built-in type `cytoscape.Css.NodeShape`

2. We need to keep `CytoscapeRenderer` as a custom type because:
   - When `cytoscape("renderer", "base")` is called, it returns `unknown`
   - We need to type this return value to access `prototype.nodeShapes` safely
   - This internal renderer API isn't exposed in Cytoscape's public type definitions
   - Since we're accessing internal implementation details to extract the shapes, we need to define these types ourselves

3. The `points` field type (`number[]`) comes from the actual implementation rather than Cytoscape's type definitions, which suggest it should be a string via `cytoscape.Css.ShapePolygonPoints`. 

(I think keeping the internal implementation of `number[]` aligns with the actual usage and makes more sense than strictly following their type definition of space-separated string in this case. But I'm fully open to suggestions/feedback/concerns on the decision).

frontend/src/util/shapes.ts
AleruDivine: Thank you for the suggestion.

frontend/src/util/shapes.ts
vincerubinetti: Thanks for leaving a detailed reply. This all makes sense. I ended up taking a look at this myself, and yeah there didn't seem to be a good way to extract the types from the library.

Take a look at what I ended up doing in #41. After trying to make use of the `points` field directly from cytoscape, I decided that it was a little too messy, and the shapes didn't even look that good anyway. It turns out not all of the `points` were `number[]`, some were functions to generate points, and some were undefined.

So instead, I just made use of cytoscapes custom polygon shape feature and defined my own shapes that I liked better anyway.

PR #39, opened 2024-10-21
Implement basic interproscan visualization component

kemsguy7:
- implement basic interproscan visualization using nightingale library. install needed components, define types for custom elements. takes as inputs: a sequence and a list of tracks/rows, each with a label and list of fragments/locations with labels and types
- generate some fake data for viz on testbed


kemsguy7:
@vincerubinetti and @jananiravi , please kindly review

vincerubinetti:
@kemsguy7 I am in the process of reviewing this and developing it more. There will be a lot more work needed here to make it a fully-fledged component. Note that this is not a critique of your work; we didn't define what to do or how far to go. Feel free to consider this done from your end, and claim another outreachy issue if you want.

kemsguy7:
@vincerubinetti thanks. 
Would love to be involved in further development too. I'll continue working on other issues for the time being.

frontend/src/components/IPR.tsx
```diff
@@ -0,0 +1,128 @@
+import { Fragment, useMemo } from "react";
+import type NightingaleInterproTrack from "@nightingale-elements/nightingale-interpro-track";
```
jananiravi, line 2: domain architecture track?

frontend/src/components/IPR.tsx
```diff
@@ -0,0 +1,128 @@
...
+import type NightingaleManager from "@nightingale-elements/nightingale-manager";
+import type NightingaleNavigation from "@nightingale-elements/nightingale-navigation";
+import type NightingaleSequence from "@nightingale-elements/nightingale-sequence";
```
jananiravi, line 5: the top sequence track?

frontend/src/components/IPR.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const commonProps = {
+    length: sequence.length,
+    height: 50,
+    "display-start": 1,
+    "display-end": -1,
+    "margin-top": 0,
+    "margin-bottom": 0,
+    "margin-left": 0,
+    "margin-right": 0,
+  };
```
jananiravi, line 46: Is the sequence derived from the top accession number in the list of accession numbers with domain architectures? Will this be protein amino acid sequences instead of current nucleic acid ones? @epbrenner better to keep this track or remove for the domarch figure?

frontend/src/pages/Testbed.tsx
```diff
@@ -122,6 +124,41 @@ for (let times = 0; times < 10; times++) {
...
+  .fill(null)
+  .map(() => ({
+    label: sample(["Lbl.", "Label", "Long Label", "Really Long Label"]),
+    features: Array(random(1, 3))
+      .fill(null)
+      .map(() => {
+        const start = random(1, Math.floor(sequence.length / 2));
+        const end = random(
+          start + Math.floor(sequence.length / 4),
+          sequence.length,
```
jananiravi, line 144: domarch lengths? length of each domain shown with those labels? The start-end coordinates will be derived from the cln_iprscan.tsv file, too.

jananiravi (commented):
Added minor comments before the testbed disappears later this week due to the 90-day cutoff! 

@epbrenner anything else to add/rm from the iprscan viz?

frontend/src/components/IPR.tsx
vincerubinetti: This component knows nothing about the biology or science. It just takes in a string (doesn't even have to be G/T/A/C, though we could constrain the type to that), and a list of items that have a label and type and are associated with a start/stop point in the sequence.

frontend/src/pages/Testbed.tsx
vincerubinetti: This is just fake data, but if there is a change I can make to make it look more like the eventual real data, I can make it.

vincerubinetti: Does this mean leave the PR open until then? Or merge it and revisit it when we have real data?

The intention here is to just establish some base visualization features -- e.g. we want to provide an arbitrary sequence, show it with zoom/scroll capabilities, then show some colored blobs with labels and detailed tooltips below and horizontally aligned with the sequence  -- then come back to tweak as necessary.

jananiravi: Let's merge since your initial PRs are all w/ mock data. We will definitely revisit them all, tab-by-tab, with real data once this initial iteration is done. We can also collect as much feedback as possible on these tabs via GH issues/discussions so we're ready to implement with real datasets.

PR #40, opened 2024-10-27
updated: added lighthouse testing functionality

### LINKED ISSUE #33: Integrate Lighthouse testing
See this related issue about color contrast which is the motivation behind this issue:
https://github.com/JRaviLab/molevolvr2.0/issues/22

As a habit, I periodically check webapps I develop with Chrome Devtool's built-in Lighthouse checks. However, I've never tried running Lighthouse in an automatic fashion, and exhaustively for every page. I'd like to at least try it in this app.

Here is the package we'd most likely use:
https://www.npmjs.com/package/playwright-lighthouse

Some notes/guidelines:
- Rename `accessibility.spec.ts` to `axe.spec.ts`
- Create a new `lighthouse.spec.ts` next to it
- Export `paths` from the axe test, and re-use it for the lighthouse test
- Choose some sensible thresholds for the lighthouse tests (90 for accessibility, 80 for performance, etc).
- Configure Lighthouse to use the "experimental APCA" option, assuming that option exists in the package as it does in Chrome Devtools: Screenshot 2024-10-15 at 12 52 39 PM
- Reconfigure Playwright to run a _production_ build of the web app instead of a _development_ build, because things in the dev build like HMR penalize Lighthouse scores like performance.

Note that this would _not_ be a replacement for the Axe Core testing that is already in place:
https://github.com/GoogleChrome/lighthouse/issues/15215
If I had to only pick one, Axe Core is much more useful.

kemsguy7:
Closes #33 

- upload test reports when running on gh-actions
- upgrade all packages
- change playwright config to use more cpu cores/workers
- change playwright server command to build production version of site
- hoist list of `paths` (routes) into separate file
- in axe test, instead of `console.warn`, use playwright test annotation (shows when you click on a test that passed in the html report)
- change nature of json imports in tests to get rid of node log warnings
- implement lighthouse tests following the [`playwright-lighthouse` instructions here](https://github.com/abhinaba-ghosh/playwright-lighthouse?tab=readme-ov-file#usage-with-playwright-test-runner)
- choose appropriate thresholds for lighthouse test, except for performance which is held back by [some kind of upstream bug](https://github.com/abhinaba-ghosh/playwright-lighthouse/issues/31)
- APCA color contrast checking is unfortunately [NOT available in lighthouse](https://github.com/GoogleChrome/lighthouse/discussions/16237#discussioncomment-11091956)


kemsguy7:
@vincerubinetti, please kindly review and leave feedback, thanks

frontend/package.json
```diff
@@ -11,7 +11,8 @@
     "test:e2e": "playwright test",
     "test": "bun run test:lint && bun run test:types && bun run test:e2e",
     "test:spelling": "bunx cspell src/**/*.{tsx,ts}",
-    "clean": "rm -rf node_modules dist bun.lockb && bun pm cache rm"
+    "clean": "rm -rf node_modules dist bun.lockb && bun pm cache rm",
+    "test:lighthouse": "playwright test tests/lighthouse.spec.ts"
```
vincerubinetti: I don't think this is really necessary. If we were going to add it, it'd make sense to add another script for Axe as well.

But you can just run e.g. `bun run test:e2e lighthouse`, using [Playwright's path argument](https://playwright.dev/docs/test-cli#reference), which is almost as short and more flexible.

frontend/package.json
```diff
@@ -70,8 +71,13 @@
     "eslint-plugin-prettier": "^5.2.1",
     "eslint-plugin-react-hooks": "^4.6.2",
     "eslint-plugin-react-refresh": "^0.4.10",
+    "fs": "^0.0.1-security",
```
vincerubinetti: This package and path should not be installed, they are not the packages you think they are. Fs and path are built into Node/Bun.

frontend/package.json
```diff
@@ -70,8 +71,13 @@
...
     "globals": "^15.9.0",
+    "lighthouse": "^12.2.1",
     "msw": "^2.3.5",
+    "path": "^0.12.7",
+    "playwright": "^1.48.2",
```
vincerubinetti: This is a shortcoming of the `playwright-lighthouse` readme documentation, but this `playwright` installation is not needed because it is already installed as a part of `@playwright/test`.

frontend/playwright.config.ts
```diff
@@ -5,6 +5,7 @@ const url = `http://localhost:${port}`;
 
 export default defineConfig({
   testDir: "./tests",
+  //testMatch: "tests/lighthouse.spec.ts", //This runs light house test for only the lighhouse.spec.ts file
```
vincerubinetti: Please take care to be consistent with the rest of the codebase, which uses lower-cased JS-doc style comments.

frontend/playwright.config.ts
```diff
@@ -16,12 +17,12 @@ export default defineConfig({
...
     { name: "webkit", use: { ...devices["Desktop Safari"] } },
     { name: "firefox", use: { ...devices["Desktop Firefox"] } },
-    // { name: "Mobile Chrome", use: { ...devices["Pixel 5"] } },
-    // { name: "Mobile Safari", use: { ...devices["iPhone 12"] } },
   ],
 
   webServer: {
-    command: `bun run dev --port ${port}`,
+    command: process.env.CI
+      ? `bun run build && bun run start --port ${port}` // production build for CI/Lighthouse
```
vincerubinetti: There is no script in `package.json` named `start`. I think you were referring to `dev`, but that is also not correct because that runs a dev-mode version of the app. What you want here is `bun run build` (builds a production version of the app) then `bun run preview` (serves the built/bundled app).

frontend/playwright.config.ts
```diff
@@ -16,12 +17,12 @@ export default defineConfig({
...
+      : `bun run dev --port ${port}`, // dev build for regular testing
```
vincerubinetti: I don't see a real reason to have the behavior here differ based on whether it's running on CI or not.

frontend/tests/axe.spec.ts
```diff
@@ -1,14 +1,15 @@
...
 log();
 
-/** pages to test */
-const paths = [
+/** pages to test */ 
+//exported path Also to be used in lighthouse.spec.ts
+export const paths = [
   "/testbed",
-  "/",
+  "/",
```
vincerubinetti: Please take care to lint/format your code.

frontend/tests/axe.spec.ts
```diff
@@ -1,14 +1,15 @@
...
 import analyses from "@/fixtures/analyses.json" with { type: "json" };
 import { log } from "./util";
 
 log();
 
-/** pages to test */
-const paths = [
+/** pages to test */ 
+//exported path Also to be used in lighthouse.spec.ts
+export const paths = [
```
vincerubinetti: Instead of duplicating this between the Axe and Lighthouse test, you can just extract the paths to a separate `paths.ts` file, and import the paths from both tests. The error you got from importing the paths directly from the Axe test file is probably a limitation of the Playwright testing suite. You probably can't cross reference objects or any other imports between test files. But hoisting the paths list into a separate file seems to solve it.

frontend/tests/lighthouse.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+};
+
+const config = {
+  extends: "lighthouse:default",
+  settings: {
+    formFactor: "desktop" as const,
+    screenEmulation: { disabled: true },
+    //emulatedUserAgent: "desktop",
+    // Enable APCA contrast checking
+    accessibilityScoring: "apca",
```
vincerubinetti: Where did you find this option? I could not find a reference to `accessibilityScoring` in the `playwright-lighthouse`, `lighthouse`, or `axe-core` source code, or anywhere on the internet in fact.

frontend/tests/lighthouse.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+  },
+};
+
+const port = 1234;
+
+// Ensure reports directory exists
+const reportsDir = "lighthouse-reports";
+await mkdir(reportsDir, { recursive: true });
```
vincerubinetti: This shouldn't be needed. Typically this type of low level thing would be handled by the library exporting the file. I tested removing this line and starting with no folder, and the `playwright-lighthouse` library was able to create the folder itself.

frontend/tests/lighthouse.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+// Generic page lighthouse test
```
vincerubinetti: Please take care to be consistent with the rest of the code base, which uses lower case JSDoc style comments.

frontend/tests/lighthouse.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+const checkPage = (path: string) =>
+  test(`Lighthouse check ${path}`, async ({ browser }) => {
+    test.skip(
+      browser.browserType().name() !== "chromium",
```
vincerubinetti: You can just get `browserName` from the function argument, as the axe test does.

frontend/tests/lighthouse.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    const context = await browser.newContext({
+      viewport: { width: 1280, height: 720 },
+    });
+    const page = await context.newPage();
+
+    // Sanitize path for filename
+    const sanitizedPath = path.replace(/[\/\?#]/g, "-").replace(/^-|-$/g, "");
+    const reportName = `lighthouse-${sanitizedPath}`;
+    const reportPath = join(reportsDir, reportName);
+    console.log(reportPath);
```
vincerubinetti: Instead of these console logs, it'd be more helpful to just look at the full report that Lighthouse produces. I added a step to the GitHub Actions workflow that uploads the report as an artifact so we can inspect tests run on CI.

frontend/tests/lighthouse.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+  "/testbed",
+  "/", 
+  "/load-analysis", 
+  "/new-analysis",
+  "/about",
+  ...analyses.map((analysis) => `/analysis/${analysis.id}`),
+];
+
+const thresholds = {
+  performance: 80,
```
vincerubinetti: I unfortunately have found the performance tests to be inaccurate using this library, and as such have reduced this threshold by a lot:

https://github.com/abhinaba-ghosh/playwright-lighthouse/issues/31

frontend/tsconfig.node.json
```diff
@@ -1,15 +1,16 @@
...
   "compilerOptions": {
+    "target": "es2017",
     "allowSyntheticDefaultImports": true,
     "composite": true,
     "module": "ESNext",
     "moduleResolution": "bundler",
     "resolveJsonModule": true,
     "skipLibCheck": true,
-    "types": ["axe-playwright"],
+    "types": ["node", "@axe-core/playwright"],
```
vincerubinetti, line 10: The "node" types I believe are included by default... Or at least, my tooling is able to recognize built-in node types without this specified here.

vincerubinetti:
I've left some comments and reworked this PR. In your screenshot, it doesn't show the actual lighthouse test results. Are you sure they were running? Before my modifications, the lighthouse tests didn't seem to be running or passing.

I'm still working on figuring out the right thresholds for the lighthouse test, and how to get the tests running faster on CI and not timing out.

PR #43, opened 2024-11-15
Download network including legend, misc fixes

vincerubinetti:
- replace `react-to-text` with more maintained `react-children-utilities`
- add `dom-to-image-more` package for "screenshotting" DOM elements
- refactor network code a bit for readability
- have download feature include the legend, instead of just the network canvas. **this sacrifices having transparency** in downloaded PNGs, but hopefully that is an okay sacrifice. the network does react to dark mode changes, so you can make figures for dark slide backgrounds for example. 
- refactor download util funcs to separate downloading and url-ifying functionality


frontend/src/components/Network.tsx
```diff
@@ -379,6 +381,12 @@ const Network = ({ nodes: _nodes, edges: _edges }: Props) => {
     [_edges, nodes, minEdgeStrength, maxEdgeStrength, edgeColors],
   );
 
+  /** fit view to contents */
+  const fit = async () => {
+    await sleep(10);
```
falquaddoomi, line 386: Why the sleep, out of curiosity? Just giving time for things to settle down?

falquaddoomi (approved):
Nice, seems like a lot of good simplifying changes. I think not having a transparent background for the PNG export is fine, personally, especially since as you said you can get it in dark and light versions.

frontend/src/components/Network.tsx
vincerubinetti: Yes, this is a little bit hand-wavey, and it's really only needed where I'm calling it from `graph.current.on("layoutstop", fit)` because I found that some of the layout algorithms (I think the ones that are more continuous/animated) aren't truly done when that event is called.

vincerubinetti: I just pushed a change to clarify this.

Idk if it's the layout algo plugins that are lying about being truly done or if it's some shortcoming in the cytoscape core, and I didn't feel it's worth it to investigate further atm, but in testing this timeout value seems to be sufficient for now.

PR #44, opened 2024-11-15
fix view corner css, improve scroll to util func

vincerubinetti:
- make another change and improvement to scrollTo func. see [https://stackoverflow.com/questions/49318497/google-chrome-simultaneously-smooth-scrollintoview-with-more-elements-doesn for lng discussion](https://stackoverflow.com/a/78850345/2180570) for a long discussion about the chrome bug and different workarounds
- remove some vestigial css and fix view corner z-index css bug

### DIFF
--- a/frontend/src/components/TableOfContents.tsx
+++ b/frontend/src/components/TableOfContents.tsx
@@ -9,12 +9,7 @@ import {
 } from "@reactuses/core";
 import Link from "@/components/Link";
 import Tooltip from "@/components/Tooltip";
-import {
-  debouncedScrollTo,
-  firstInView,
-  isCovering,
-  scrollTo,
-} from "@/util/dom";
+import { firstInView, isCovering, scrollTo } from "@/util/dom";
 import { sleep } from "@/util/misc";
 import classes from "./TableOfContents.module.css";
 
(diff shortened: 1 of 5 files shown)


vincerubinetti:
FYI the thing to test here would be the testbed page. Scroll up and down the page to see the TOC active item stay in view with instant scroll, and then click on a TOC item to see the page smoothly scroll to it.

vincerubinetti:
@falquaddoomi poke, when you're done with 🎓

falquaddoomi (approved):
Nifty! I tried scrolling around the main page with the trackpad as well as navigating through elements with the keyboard, and both maintained the TOC highlight sync as expected. Activating the TOC to scroll to an element with a mouse click as well as with the keyboard also appeared to work well. Opening a link with a fragment (e.g., https://deploy-preview-44--molevolvr.netlify.app/testbed#popover) correctly scrolled to the element and highlighted the "Popover" TOC entry.

I tried the following (convoluted) test, too, and it worked:
1) focusing an element in the TOC, then
2) scrolling away from it in the TOC as well as on the main page, then
3) activating the no-longer-visible focused element in the TOC by pressing enter.
4) Scrolling the main page correctly highlighted the TOC entry for the element at the top of the viewport.

The code looks good, and I think the workaround for the Chrome issue you mentioned makes sense.

PR #46, opened 2025-01-21
Sunburst chart

vincerubinetti:
I started off with Plotly.js, then got frustrated with its API and limitations, then tried Echarts which was better but still didn't provide the options and simplicity I wanted, so I eventually just wrote it myself. I feel this looks better and will be more customizable, and is actually more simple/predictable for almost the same amount of code. One perhaps key thing the libraries do that my custom version doesn't (yet) is allow you to click on a slice and "expand" it. If we needed that it'd then become more code than using the libraries, but I'm not sure how useful that is.

- pull in misc small component updates from geneplexus
- in network component:
  - use reactive color map for node/edge colors like everywhere else in code base, instead of "colorMix" util function
  - fix bug where colors wouldn't update when dark mode changes
  - truncate edge labels based on distance between nodes
- add custom sunburst chart component
- add fake data for testbed test of sunburst
- refactor color palette, re-choose colors
- move around some util funcs

I made some changes to the color util funcs to make the sunburst chart look good, which led to me needing to change the network component as well.


frontend/src/components/Network.tsx
```diff
@@ -592,17 +604,8 @@ const Network = ({ nodes: _nodes, edges: _edges }: Props) => {
...
-          .toPng(root.current, { scale: 2 })
-          .then((blob) => downloadPng(blob, "network"));
-
-      if (format === "jpg")
-        domtoimage
-          // @ts-expect-error non-comprehensive types for dom-to-image-more
-          .toJpeg(root.current, { scale: 2 })
-          .then((blob) => downloadJpg(blob, "network"));
+      if (format === "png") downloadPng(root.current, "network");
+      if (format === "jpg") downloadJpg(root.current, "network");
```
jananiravi, line 608: can we add vector graphics -- svg/pdf to this and other download options?

frontend/src/components/Network.tsx
```diff
@@ -563,18 +551,42 @@ const Network = ({ nodes: _nodes, edges: _edges }: Props) => {
...
+        truncate(newNode?.label, { length: 4 * (minNodeSize / fontSize) }),
+      );
+
+      /** when node dragged */
+      node.on("position", () => {
+        node.connectedEdges().forEach((edge) => {
+          /** set truncated label based on edge length */
+          const a = edge.sourceEndpoint();
+          const b = edge.targetEndpoint();
+          const dist = Math.sqrt((a.x - b.x) ** 2 + (a.y - b.y) ** 2);
```
jananiravi, line 580: didn't realize this is the default eqn used for edges. btw, naive Q: why are these network scripts being adjusted in the sunburst PR?

frontend/src/components/Network.tsx
vincerubinetti: We'd need to use a library for this, either pdf.js or jsPDF. Both of them are [fairly heavy](https://bundlephobia.com/) packages, considering that in our case they'd only serve one narrow function. But I could have the library be only loaded on a demand. It would be some work, and I'd want to add it to all the other charts for consistency as well.

I would consider how needed this is vs. the KB of bandwidth it will use. If I want to attach a chart to an email, I'd personally just use a JPEG. If I want to make a report, I'd drag the JPEG into a word doc. When is a single page PDF desired? I guess if you wanted to print it, maybe you could want a PDF, but the user could also ctrl + p to print the page to pdf.

If you'd be okay with using the browser's print-to-pdf functionality, it might be possible to do this without a library:

- When button clicked...
- Temporarily make chart full screen size and hide everything else on page
- Open browser's print dialog with `window.print`
- Revert style changes when user done printing

frontend/src/components/Network.tsx
vincerubinetti: Not a naive question. To make the sunburst chart look nice, I had to rework the color util funcs. Because of that rework, I had to make some changes to the network component which were also using the funcs. While making these changes, I had to write this `graph.current.nodes().forEach`, which I realized was the natural place to put a `node.on("position", () =>` callback which could be used to update the edge labels.

So, no, it was not strictly related to this PR, but I threw it in rather than doing a 5 line separate PR.

Related:
https://www.youtube.com/watch?v=AbSehcT19u0

vincerubinetti:
> Qn: The legend colors are in the order of these concentric rings -- inside to outside, but can we have a 'layered legend' for better resolution?

I think you just mean having the legend be more vertical than horizontal? If not can you do a napkin sketch? I put them that way to be most compact, but I could also make them a vertical stack and put them to the left of the rings. It'd still "collapse" back to how it is now on small width screens.

frontend/src/components/Network.tsx
jananiravi: Yes, pdf's rendered within the browser should be a good start.

vincerubinetti: Note that there is a limitation with this method. JavaScript doesn't/can't run inside the print dialog. You'll see that the sunburst chart will nicely update (e.g. the legend will wrap to a new line) when you change from portrait -> landscape. That's because its layout is based on plain HTML and CSS. The network, however, being canvas-based, has to be resized with a call to the `fit()` JavaScript function to put all the nodes nicely inside the view. As such, there's not really any way we can make the network automatically and always print nicely. The user would have to make their window size roughly the aspect ratio of how they want to print, then print.

PR #47, opened 2025-02-03
React 19

vincerubinetti:
- upgrade all packages. major version bumps: [reactuses/core](https://www.reactuse.com/changelog#breaking-changes-1), [react](https://react.dev/blog/2024/12/05/react-19), [vite](https://vite.dev/blog/announcing-vite6), [react-router](https://reactrouter.com/upgrading/v6)
- for react-router, since we're still using it just "as a library" instead of "as a framework", nothing has changed except a few import names
- for react 19, much has changed, but in our case it mostly just means `forwardRef` is no longer needed, and some typing tweaks were needed.
- for other libraries, no breaking changes that affect us.
- replace a few "sleep().then()" calls with useEffects which should achieve same thing in more idiomatic way (call side-effect outside of render)


frontend/src/components/Radios.tsx
```diff
@@ -67,8 +66,10 @@ const Radios = <O extends Option>({
 
   /** notify parent when selected changes */
   const previousSelected = usePrevious(selectedWFallback);
-  if (previousSelected && previousSelected !== selectedWFallback)
-    sleep().then(() => onChange?.(selectedWFallback));
+  useEffect(() => {
+    if (previousSelected && previousSelected !== selectedWFallback)
+      onChange?.(selectedWFallback);
+  }, [selectedWFallback, previousSelected, onChange]);
```
falquaddoomi, line 72: 🎉

falquaddoomi (approved):
I've been out of React development for a while, but I assume getting rid of `sleep()` calls in favor of events is a good thing, and you've done that a lot in this PR, in addition to upgrading React. Nice work!

PR #48, opened 2025-02-03
Dialog component

### LINKED ISSUE #45: Dialog component
I tend to dislike dialog modals, especially when they take up a big part of the screen. But we'll probably need one eventually (to hopefully be used very sparingly).

Should be able to use largely the same code as the Popover component.

vincerubinetti:
Closes #45 

- fix dark-mode FOUC by subbing to state atom directly 
- add dialog component with content slot and actions/bottom slot
- add dialog component to testbed
- global style tweaks to support dialog component


falquaddoomi (commented):
Sorry for the delay in reviewing this. Cool way to fix the FOUC, and the dialog component looks good!

I noticed a few things:
- Attempting to select an entry from the dropdown that extends past the bottom of the dialog dismisses it.
- No worries if this is by design, but I also noticed that the left and right arrow keys scroll through the dropdown options rather than up and down as I'd expect. Down arrow appears to expand the list, and I wasn't able to find a key that collapses it again.

vincerubinetti:
Good catch, I didn't notice this. Up/down arrows should be able to change the selection in the Select components. The right/left arrows are conveniences I added and only apply when the select is focused but not opened. Pressing escape while the dropdown is open should close just the dropdown, not the dialog.

Both of those are due to the same issue of nesting the select inside the dialog, which seems to be a known and frustrating issue in the Radix UI library. It's made more difficult by the fact that I'm using Dialog from Radix UI and Selects from Headless UI (because Radix doesn't support multi-selects).

I will look into this. Might just end up using Headless UI's Dialog, assuming it will work better with the Selects. But then if we end up using any other library's components in a dialog at some point, we may end up with the same problem.

vincerubinetti:
Okay I've made a fix. I don't really like it but after thoroughly exploring workarounds, params, and other libraries (e.g. RAC), this is the best I can come up with for now. I'm using Headless UI (HUI) Dialog now and it seems to play well with the HUI Selects, and this should be the only component (for now) that can "break out" of the dialog. I wanted to also use HUI for the Popover component since it's very similar, but ran into issues (no focus trapping, no inerting bg elements). Keeping the Radix UI Popover, there is an issue: pressing escape on an open dropdown will close both the dropdown and the popover (not the case with the Dialog component). But I feel this is okay enough because the user can still tab out of the Select to exit it.

falquaddoomi:
Heh, I can appreciate your disappointment with these libraries, but whatever you've figured out is working well for me, at least! I tried the updated version of the component on the Netlify preview, and it seems to work exactly as I'd expect: I can select things from the dropdown with the mouse, the keyboard shortcuts are IMO more intuitive, and escape exits out of things in intuitive order.

As for the dropdown in the popover, like you said the user can work around it with other keys, and IMHO if it were to become a problem I suppose we could use a dialog instead. Nice work!

falquaddoomi (approved):

PR #49, opened 2025-02-11
Refactor sunburst

vincerubinetti:
- match svg font size to document font size
- refactor using d3 util funcs, significantly reducing LOC and complexity. much less custom math, no more recursion (because d3 tree structure gives nice methods to access what we need), more flexible
- use different method for vertical centering of text (thanks to safari not supporting `dominant-baseline` with `text-path`)
- update testbed to show more children slices, and sometimes have only 1 root slice (testing full circle, which is edge case with SVG arc commands)
- update viewbox fit util func to return individual numbers
- write generalized hook to adapt font size (or any other size) in an svg to the size of something in the document


vincerubinetti:
@falquaddoomi poke for code review here. This is almost a pure refactor, meaning if you compare it to the current pie chart on `main` it should be the same (other than some style differences like font size).

falquaddoomi (approved):
Sorry for the huge delay in reviewing this. I did a side-by-side comparison with the old version, and it's working pretty much the same as the old one was. I tried all the export format options, tried keyboard navigation, and resizing the viewport in various ways; it all works as I'd expect it to.

I glanced over the code and it is indeed significantly less complex; kudos!

PR #50, opened 2025-03-06
Msa

vincerubinetti:
- switch to `html-to-image` (better maintained fork of `dom-to-image` than `dom-to-image-more`)
- make basic MSA component
    - takes multiple string sequences, with labels
    - assigns characters in sequence a particular color based on a "type" (provided by consumer or auto-assigned by uniq character)
    - combined summary row at top that shows percentage breakdown of chars in column
    - ability to wrap to new rows at N number of chars
    - print and raw export options
- debounce network component resizing that was slightly slowing down page
- when clicking number box inc/dec buttons, keep button under mouse
- split up testbed page code. split each section into its own component, making it easier to re-arrange and comment/uncomment them. split fake data generation functions into separate file.
- change color palette yet again. go back to hand-picked tailwind colors.
- tweak some util funcs

I'll shortly also be adding JPEG/PNG export to the MSA viz as well. I won't be adding SVG export, because the way it's currently constructed using several separate SVGs. I could reconstruct the whole viz as one large SVG, which would have the following tradeoffs: allow single SVG download, more straightforward but much more verbose positioning of elements, less flexible and harder to make changes to layout, not easily made responsive to screen size width.


falquaddoomi (commented):
The graphic looks great IMO! I tried out the following features:
- **Keyboard navigation:** I can tab through each panel and scroll them with the arrow keys as expected.
- **Wrapping:** seems to work as expected, with additional rows being added as the wrap width is decreased. The window does indeed remain scrolled so that the mouse is under the button you're clicking, even as the height of the page is changing, which is pretty cool. :)
- **Exports:**
  - the PNG and JPEG exports appear to export the image as it appears in the page. One thing I noticed is that the PNG and JPEG previews appear to show the current view, but the panel in the resulting image is scrolled to 0, rather than showing the portion of the panel where the user had scrolled. *(More thoughts in [1].)*
  - The TSV export works well, although if it's not a standard format you might consider adding a header column.
  - The PDF export works, and IMO the result looks great.

*[1] I don't know how big of a problem this is, if at all; I also don't know if there's a way to get the package you're using to manipulate the scroll position in the exported image. If there isn't, you could consider setting the wrap prior to export to a value for which the contents of the panel won't overflow and require scrolling.*

vincerubinetti:
> although if it's not a standard format you might consider adding a header column.

I think in this case column headers aren't super useful? After the first column, it'd just be the sequence position. I guess I'd leave that call up to Janani & Co. It would just look something like:

```
Name,1,2,3,4,5,6,7,8,9,10,
Firmucutes A,A,G,T,C,C,T,A,T,G,
Firmucutes B,A,G,T,C,C,T,A,T,G,
```

Maybe it could be more useful if we include the top "combined" row to the tsv export. Not sure how to format that... just choose the most common character in each column? show all the %s like `A 70%\nG 30%` (or comma separated).

vincerubinetti:
Regarding the jpg/png export issue:

I didn't notice that scroll issue. I would've assumed that the library would preserve whatever was on screen. It looks like this is a known issue: https://github.com/bubkoo/html-to-image/issues?q=is%3Aissue%20state%3Aopen%20scroll%20position I'm not sure the library could even fix the issue though, as I think it would lie in the browser's internal painting of the elements to canvas. One workaround is instead of centering a particular element in the view using the browser's normal overflow, you'd (temporarily) hard code a fixed offset position of everything in the container with CSS, as described in some of those issues. That's very dirty though and I'd like to avoid it.

Also, my thought is that if a user wants something to be visible in the png and were going to manually scroll to it anyway, they could also manually set the wrap point to ensure it's visible without a scrollbar at all.

Another option is to temporarily automatically set the wrap point ourselves right before the user exports as png, then reset it afterward. This is what I'm doing with the print option. But in the case of printing, we can usually expect the screen size to be a standard US letter, pretty wide. But with the png, it depends on the user's window width. I'll try experimenting with this. An exact calculation of the wrap point would be complex, but I could either loosely base it on window width, or iteratively reduce it until there's no scrollbars.

vincerubinetti:
> could either loosely base it on window width, or iteratively reduce it until there's no scrollbars.

The iterative can work, but causes a visible lag because you need to decrease the wrap, wait for it to render, check if a scrollbar is there, decrease again, etc. The rough tuning seems to work well enough, though when CSS on the page changes (widths of stuff) in the future, a maintainer will have to remember to re-tune it.

falquaddoomi:
> > although if it's not a standard format you might consider adding a header column.
> 
> I think in this case column headers aren't super useful? After the first column, it'd just be the sequence position. I guess I'd leave that call up to Janani & Co. It would just look something like:
> 
> ```
> Name,1,2,3,4,5,6,7,8,9,10,
> Firmucutes A,A,G,T,C,C,T,A,T,G,
> Firmucutes B,A,G,T,C,C,T,A,T,G,
> ```
> 
> Maybe it could be more useful if we include the top "combined" row to the tsv export. Not sure how to format that... just choose the most common character in each column? show all the %s like `A 70%\nG 30%` (or comma separated).

Well, I was just thinking it'd be "Name\tSequence", just so it's identified somehow; no need to create columns per position in my opinion. That said, if the version without the header is an already-accepted format, let's stick with that.

vincerubinetti:
> if the version without the header is an already-accepted format

I have no idea about this. @jananiravi @epbrenner, is there a particular format that's most standard or helpful when downloading as tsv? See [my comment above](https://github.com/JRaviLab/molevolvr2.0/pull/50#issuecomment-2718849049)

vincerubinetti:
Also @jananiravi @epbrenner regarding the [clustal color scheme](https://www.jalview.org/help/html/colourSchemes/clustal.html), all you'd have to do is use the component like this:

```
const msaTracks = [
    {
        "label": "Some Label",
        "sequence": "XPWHCZ..."
    },
    // etc
];

// map sequence characters to types
const types = {
  "A": "Hydrophobic",
  "C": "Hydrophobic",
  "E": "Negative charge",
  // etc
};

<MSA tracks={msaTracks} types={types} />
```

Inside the component, the `types` map is used to map sequence characters to arbitrary types, which are then auto-mapped to our color scheme (which is carefully chosen to have good contrast, look nice, and work in light/dark modes).

If you'd like me to allow the component to show a legend for the colors, please let me know. If tooltips one certain elements would be helpful, we can add those too (though adding them to every sequence character could impact page performance a bit).

falquaddoomi:
> Another option is to temporarily automatically set the wrap point ourselves right before the user exports as png, then reset it afterward. This is what I'm doing with the print option. But in the case of printing, we can usually expect the screen size to be a standard US letter, pretty wide. But with the png, it depends on the user's window width. I'll try experimenting with this. An exact calculation of the wrap point would be complex, but I could either loosely base it on window width, or iteratively reduce it until there's no scrollbars.

If that's what you've implemented in your most recent commit, I just tried it and it seems to be working well; I can see the entirety of the sequences in the resulting PNG. There's a brief flash of it changing the wrapping, of course, but that's IMO totally fine.

A very minor thing, but if the width toggle is in "wide" view the resulting PNG/JPEG ends up being as wide as the screen, but the contents are still constrained to it being in the narrow view (I think?)

Here's an example with the width toggle set to "wide":

<kbd>MolEvolvR_msa (4)</kbd>

And just for comparison, with it set to "narrow" (what I assume is the correct behavior):

<kbd>MolEvolvR_msa</kbd>

vincerubinetti:
@falquaddoomi good catch, the latest commit should improve the spacing there (but still not perfect because it's just a rough tuning).

epbrenner:
> > if the version without the header is an already-accepted format
> 
> I have no idea about this. @jananiravi @epbrenner, is there a particular format that's most standard or helpful when downloading as tsv? See [my comment above](https://github.com/JRaviLab/molevolvr2.0/pull/50#issuecomment-2718849049)

People going to download MSAs are usually expecting one of a few possible plaintext formats. These are usually Clustal, FASTA/Pearson, and _sometimes_ NEXUS. If somebody wants to download the MSA not in a graphical format, they'll probably want it in a format like that so they can use it in other software downstream (like building their own custom phylogenetic trees). 

The raw input MSA that goes into this is the simplest plaintext output to provide a user, but if they want the `Combined` consensus sequence, that's something not present in that original. Maybe the options for plaintext/tsv download could be "Original MSA" in the original upstream format MolEvolvR produces before visualization, and then "Consensus sequence" just as a FASTA file like:


```
>ConsensusSeq_$QueryName_$FiltersApplied
MWYFAWILTSTLWARE[...]

```

And this separate file would contain only the combined consensus sequence as a regular FASTA format sequence and no other sequences. At least that's what I think may be the most broadly applicable, but I'd like thoughts from @jananiravi too!

falquaddoomi:
> @falquaddoomi good catch, the latest commit should improve the spacing there (but still not perfect because it's just a rough tuning).

@vincerubinetti: Your latest change seems to have fixed it for me!

I haven't noticed much roughness myself, but IMO whoever's downloading it can edit to their liking, if their intent is to use it for a publication or something. If they want a more predictable output format they can download it as a PDF, and for a more flexible format there's the TSV.

falquaddoomi (approved):

vincerubinetti:
@falquaddoomi After you approved, I noticed some critical bugs I wanted to include, and also ended up including a legend and the clustal coloring scheme that was requested. I'm also still waiting to hear back about what TSV download format is desired, which I'll also add to this PR.

@epbrenner Could you take a look at the code in `msa-clustal.ts` to verify that I've interpreted the table correctly? And please spot check that the coloring is right on the testbed, where I've hooked up the clustal theme by default.

jananiravi:
> I think in this case column headers aren't super useful? After the first column, it'd just be the sequence position. I guess I'd leave that call up to Janani & Co. It would just look something like:
> 
> ```
> Name,1,2,3,4,5,6,7,8,9,10,
> Firmucutes A,A,G,T,C,C,T,A,T,G,
> Firmucutes B,A,G,T,C,C,T,A,T,G,
> ```
> 
> Maybe it could be more useful if we include the top "combined" row to the tsv export. Not sure how to format that... just choose the most common character in each column? show all the %s like `A 70%\nG 30%` (or comma separated).

The MSA would be in a fasta format -- it will work later w/ any of these tools: https://www.ebi.ac.uk/jdispatcher/msa

Reg. @falquaddoomi's comment:
>Well, I was just thinking it'd be "Name\tSequence", just so it's identified somehow; no need to create columns per position in my opinion. That said, if the version without the header is an already-accepted format, let's stick with that.

I agree that for a simple tsv, this should suffice
Name\tAccNum\tSequence

Reg. @epbrenner's comment:
yes, the most common use case is exporting the MSA in FASTA format so they can use it with a different tool downstream.

Was the MSA coloring part of a different PR or resolved comments above? I know Evan had some thoughts on those -- can't find that anymore.

vincerubinetti:
> Was the MSA coloring part of a different PR or resolved comments above? I know Evan had some thoughts on those -- can't find that anymore.

If you're referring to the clustal coloring, that is implemented and in the preview.

> I agree that for a simple tsv, this should suffice
Name\tAccNum\tSequence

I've just made changes to download in this format. You can put label/name, sequence, and any additional field (not just accnumber) as a column in the download.

falquaddoomi:
Hey @vincerubinetti, as @jananiravi mentioned above, in our prior meeting @epbrenner mentioned that having FASTA as an additional download type could be useful for downstream analyses. Sorry that I'm just mentioning this now, and hopefully it's not too hard to add.

FASTA is a text format in which each sequence is preceded by a "header" line. The header starts with "> " and then lines that follow it are the sequence for that header. Multiple sequences can be included in one file, each initiated by a header line.  AFAIK it's common for a FASTA header to start with the accession number, but I'm unaware of specifics beyond that; maybe @jananiravi and @epbrenner can provide some advice.

In our case, it might look like
```
> esse velit nulla deserunt eu ut eu laborum nostrud
APLTMWOCAIUOTATCIRDOUEXEMFZRKDNKOLPGEYMBLSGNTIKQPFIADONYWUHFXJKQOZHXMDJVPTJZCKYW
PICOMLUFMBNJATTRIHMTRCWXSLBUUYFQFQOKXPVKOPQKALGWRFOFOFQWKJEHXIGENMNAIXYQAVELZXDU
WDZLTKXINRUNDVCBMXNAUIVGFVFZCHTSKSFVTQWUUMSKNPXUSSKLHOXYIZCTXOTZRJKVINROJKQDCOWQ
TPLBVOZOIRJJJRPVLOJBQMYXKRXGMOLRQNGFZCHCCOPJDIMAEJLJVHZHESWVLEHVFQEDOJOLXGEWPCIG
OPOELDZNTPATYUDIIYBAIRIMYMUQMDXKKCGVOPFFBQHCOCEAAPPAASCNNCCMCJQMYVGGZCTCZMRUNFIF
BGCPWXPTONMBXKJZOORFVZVBJKZFOBPSLMERPWXYWHLMQJOLQKSNKDOCJMPSPFPELJCCRNYYEEMSOSZR
JHKWZOFKZLGIYOMNVNDLEPCTCXDRRTBVUSMAFGWIINPZUSNZEFECWQESCBQKPLETADUTHMWVNUJFTYPJ
KMQTNGCGNHSJQBUMIZUVHTMQYZDMFRPXDNMIKHQHKTCLBINOLXLXRHVJQMKULDEVSEBJOOGSSSTUBDMP
RALSPAYWXFMYEPLOQLAKFNHSTPGKLKZCXGZIRANMAKRWYATIKBYVPAZXZRZITHOEMXJZYEQFKDUPWPBB
NNHNTMEESPFV

> nostrud nisi tempor amet incididunt qui est est est in
APLJMWOCAOUUTWRCIYDLUEXEMVZRGDNKWLBGESMBLSGNTIKQPFIADONYWUHFXUKQOZHXMDJVPTJACJYW
PICOMLULMXNJATTRIHMMRCEXSLBUUYFQFZOKXPVKOPQKALGWRFOFOFQWKJEHXIGENMPAIXYQASELZXDU
WDZLTKXINPUNDVCOBMXNAUIMGAVFZCHTSHSFVTQWUUMSKNPXUSSNLHOXYWECTXQTZZJKVINROIAQDCOW
QTPIBVOGOIPOJJRPVLOJBQMYXKEXGMOVRQNGFNCHCCPJZIMAEJLJVHZCEOWVLESVFQEDOJOLXGEWCIGO
POELDZNTPBTYUDIIYBAIRIMYMUQGBXGKCGVOPFFBQHCMCEAAPPAASCNNCSRMCJQMYVGBZCTCEMPGNFIF
BGPBOPTFVMBXKJZOORFVZVBJKZFOBPSLWERPWXPIHLMQJOLOKSZKDOCFMJSPFTELJCCRNYYEEMQOSZOJ
QKWZOFKZLGYYOMNJNONEPCTCXDRRTBVUSMEFGWLINPZUSNFEFECWEINBQMPLITADUTHMOVNUJZTYPJKM
QTNGCGNHSJRBULIZUVHTHYQYZDMFRPXDIMIKHQHKOCLBDNOLNLXRHNJMMQULDEVSEBJLOGSSTTUBDMQR
LSOYWXFMYEPLOWQLAKFNHSTYGKLKZCXMZICAIMABRWYATIKFYVPAZXZRZITOEMXJZYWQKDUPWFPVNHNT
MEESPFV
```

epbrenner:
Sorry again to keep you waiting on this @vincerubinetti! This looks good and ready to go. Will approve shortly.

The only (non-blocker) quirk I really see is when downloading the PDF in Dark Mode, the MSA goes to a white background (which I think is fine and a reasonable expected behavior), but it still retains the black background on amino acid positions that don't have a coloring assigned. Since you're using non-standard characters like J, this behavior isn't really an issue for them since you'll never see those in a biological sequence, **but for gaps (-),** these won't have properties so they'll have that black background normally. We _could_ keep that, but I also think just going with the light mode white background for gaps would work well too. 

tl;dr: The gap coloring is usually a lack of coloring because there's nothing there. The PDF gaps from a dark mode download are black against a white page background. That could be switched to white. Low priority.


Screenshot 2025-04-15 at 3 19 02 PM

epbrenner (approved):
Downloads and coloring logic all check out to me. Minor dark/light mode PDF download issue left in comments won't block merge. @ me if you'd like me to make an issue out of it to keep track of it for later!

vincerubinetti:
Regarding the print background, I've changed it so it will respect the dark mode black background. But, since we're just using the browser's built in print dialog for this, the user will still have to allow background images/colors to be drawn. In Chrome, you have to manually check this option under "More settings", and there is no way we (via JavaScript) can check it for them:

Screenshot 2025-04-15 at 7 23 47 PM

PR #52, opened 2025-05-12
Heatmap component

### LINKED ISSUE #37: Heatmap plot visualization
Please research library options and discuss them here before starting implementation. If at all possible, I want to use a library that can handle all the general plots we may need, including: bar chart, line chart, upset plot, sunburst plot, heatmap (see the other issues in this repo).

More specific notes/guidelines:

- Should be SVG-based, not canvas-based
- Don't necessarily need exactly the same feature set as the 1.0 visualization. Layout can be slightly different, can have more/less/different customization options, etc.
- Don't implement a download button yet.

See the following comment for guidelines on implementation and general architecture:
https://github.com/JRaviLab/molevolvr2.0/pull/32#issuecomment-2408741783

vincerubinetti:
Closes #37

### DIFF
--- a/frontend/src/components/Gradient.tsx
+++ b/frontend/src/components/Gradient.tsx
@@ -1,4 +1,4 @@
-import { useId } from "react";
+import { useId, type ComponentProps } from "react";
 import * as d3 from "d3";
 import { range } from "lodash";
 import type { Option } from "@/components/SelectSingle";
@@ -6,21 +6,25 @@ import type { Option } from "@/components/SelectSingle";
 type Props = {
   id: Id;
   flip?: boolean;
-};
+  direction?: "horizontal" | "vertical";
+} & ComponentProps<"svg">;
 
-/** gradient preview thumbnail */
-export const GradientThumb = ({ id, flip = false }: Props) => {
-  const filter = useId();
+/** gradient */
+export const Gradient = ({
+  id,
+  flip = false,
+  direction = "horizontal",
+  ...props
+}: Props) => {
+  const gradientId = useId();
 
   return (
-    <svg
-      viewBox="0 0 10 10"
-      width="3em"
-      height="1em"
-      preserveAspectRatio="none"
-    >
+    <svg viewBox="0 0 10 10" preserveAspectRatio="none" {...props}>
       <defs>
-        <linearGradient id={filter}>
+        <linearGradient
+          id={gradientId}
+          gradientTransform={direction === "vertical" ? "rotate(90)" : ""}
+        >
           {range(0, 1, 0.1)
             .concat([1])
             .map((percent, index) => (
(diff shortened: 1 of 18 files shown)


falquaddoomi (commented):
Nice work! The heatmap looks good, and the following functions all seem to work as expected in Chrome/Firefox on OS X:
- hovering over a heatmap cell shows its value and metadata
- hovering over an axis label displays the full label text
- gradient, flip, transpose controls work as expected
- tabbing through both cells and labels works as expected
- tabbing through and interacting with controls using the keyboard works, including help tooltips
- all the download options work (kudos for making the PNGs alpha-transparent)

One extremely minor thing is that the PDF download produces two pages on OS X in Chrome/Firefox, one with the plot and one blank page.

Safari works aside from the following issues (Version 18.4 (20621.1.15.11.10)):
- the PDF download is just one page, but it seems the figure isn't centered in the page and much of it is clipped, including the gradient scale:
[Testbed | MolEvolvR.pdf](https://github.com/user-attachments/files/20174870/Testbed.MolEvolvR.pdf)
- tabbing through the cells and axis labels work, but least in the Netlify preview I can't seem to reach the controls via tabbing; they don't appear to receive keyboard focus. The gradient dropdown can be navigated via the keyboard once expanded, though.

vincerubinetti:
> * the PDF download is just one page, but it seems the figure isn't centered in the page and much of it is clipped, including the gradient scale:
>   [Testbed | MolEvolvR.pdf](https://github.com/user-attachments/files/20174870/Testbed.MolEvolvR.pdf)

Trying to fix this, but it's proving difficult. A style change to `.print-element {` will fix the heatmap for Safari, but mess up the Network PDF download in Firefox for example. I'll keep trying but this might have to stay as a "best effort" kind of thing.

> * tabbing through the cells and axis labels work, but least in the Netlify preview I can't seem to reach the controls via tabbing; they don't appear to receive keyboard focus. The gradient dropdown can be navigated via the keyboard once expanded, though.

This is due to this unfortunate default behavior on MacOS: https://stackoverflow.com/a/67901876/2180570

More "regular" controls like inputs can be focused with tab by default, but not buttons and other things. And since our custom components are ultimately button elements in the DOM, this happens.

vincerubinetti:
I've made some changes that should improve Safari printing a bit, and not mess up any other figure PDF printing in other browsers (I've checked each manually). It genuinely seems that Safari does not respect/understand viewport CSS units when printing, so idk how much I'm able to fix here.

falquaddoomi:
It seems like what you changed fixed multiple pages appearing in Chrome/Firefox PDF downloads, and it definitely looks better in Safari! 🎉 

Thanks for letting me know about the tabbing fix for Safari; frankly, I don't normally test with Safari, so I didn't occur to me that it might be an OS/browser issue. FWIW, regarding your SO link, I ended up being able to just enable this value in the Safari settings for tabbing to work with all the controls; I didn't need to change the OS-level settings:
tabbing in safari fix

That's too bad about Safari's incomplete support for viewport units, but like you said I think a best-effort attempt with the time we have is good enough. Generally, as long as everything works in one major browser, we can make a best effort for the rest and add a disclaimer that everything might not work as well as it does in the chosen one.

falquaddoomi (approved):

jananiravi:
Appreciate the safari fix as well. The heatmap looks great! 👌🏼

PR #54, opened 2025-05-20
Tree component

### LINKED ISSUE #34: Tree viz
Example screenshot from 1.0:
Screenshot 2024-10-15 at 1 10 18 PM

Essentially needs to display a tree hierarchy of data (proteins I think), and also show a sequence visualization aligned with each protein.

I kind of doubt there will be an existing javascript library that does _exactly_ this. We'll most likely need to combine two libraries together. Perhaps we could use [echarts](https://echarts.apache.org/examples/en/index.html#chart-type-tree) for the tree part, and [Nightingale](https://ebi-webcomponents.github.io/nightingale/?path=/story/components-tracks-interpro-track--track) for the sequence part. Or maybe we can use [d3-tree](https://d3js.org/d3-hierarchy/tree) for the tree, and do the sequence part manually (it seems like it's just a sequence of colored rectangles.

See the following comment for guidelines on implementation and general architecture:
https://github.com/JRaviLab/molevolvr2.0/pull/32#issuecomment-2408741783

vincerubinetti:
Closes #34 

- consistently wrap other components in "card" styling
- focus style tweaks to other components to make consistent
- tweak fitViewBox util func params
- remove unnecessary/vestigial "title" param in sunburst component
- implement simple tree visualization with d3
- add fake tree data generation for test bed
- rearrange order of components on testbed, roughly simplest to most complex
- incorporate flushSync into useSvgTransform util func for better (in my testing) fitting behavior

### DIFF
--- a/frontend/src/components/Tree.tsx
+++ b/frontend/src/components/Tree.tsx
@@ -287,9 +287,7 @@ const Tree = ({ data }: Props) => {
             <span>To</span>
             <span>{selectedB?.data.label}</span>
             <span>Dist.</span>
-            <span>
-              {selectedDist.toFixed(2)} ({selectedPath.length} nodes)
-            </span>
+            <span>{selectedDist.toFixed(2)}</span>
           </div>
         )}
       </div>


vincerubinetti:
Reviewers, note that the component can also show details for each "intermediate" node on the tree, not just the leaf nodes at the end. Not sure if we would need that in our case, but we get it essentially for free from the architecture of the component.

Also, the structure of the data that the component takes is the same as that of the sunburst component:

```yaml
- label: "top level node"
  children:
    - label: "second level node"
      children:
        - label: "leaf node"
# etc...
```

This may not match the structure of what will come from the backend code computing the phylogeny. But in an effort to keep the component general purpose and reusable, I think it should stay this structure, and we'll eventually have a (hopefully small) bit of code to transform the phylogeny data into this.

vincerubinetti:
References

https://en.wikipedia.org/wiki/Newick_forma
https://yulab-smu.top/treedata-book/chapter7.html
https://etetoolkit.org/treeview/

vincerubinetti:
I believe this now incorporates what we discussed in the meeting yesterday.

epbrenner:
This is a excellent proof of concept for the phylogeny visualization.

I'll provide a couple minor thoughts here.

1. Measuring behavior is very cool. I notice it's also "measuring" the vertical distance (which is to say the nodes counted in Dist. include the vertical branches), but I'm guessing these vertical branch length values are 0 so it doesn't actually add anything, which is the correct way to do it. Eyeballing it, it does seem that the distance addition logic might be a little off.
Screenshot 2025-05-21 at 9 44 27 AM

This horizontal distance of 2.01 from tip-to-tip doesn't seem to make sense with...
Screenshot 2025-05-21 at 9 45 03 AM
this horizontal distance of 2.09. 

These numbers are arbitrary so I could be wrong, but the sum total of horizontal branch separating the first set of points appears much smaller than the horizontal distance separating the second set. It may be summing distance to root of two nodes vs. just the distance between selected nodes, which wouldn't require traversing all the way back to the root for the first picture.

Generally speaking, a user would typically only measure from selected tip-to-tip. In that sense, we wouldn't really _need_ the internal nodes to be their own selectable points, but those are interesting pieces of data to provide at least optionally. We may want to discuss the details of that implementation on our end to give you better guidance, but for the time being, I think only having tip nodes is sufficient, and any internal nodes just won't have the circle shapes designating a selectable node.

2. For further edge case testing, a lot of the time users might see trees where they have several tips separated by 0 branch length. These are cases where the two representative sequences are identical, so the distance between them is 0. See the green highlighting in this random tree. Most of those nodes have zero difference between them, so they sit on that vertical plank side-by-side. I'm sure your tree can handle this, but I just didn't see it occur in the randomization attempts on refreshing the page, so keep an eye out for that.
Screenshot 2025-05-21 at 9 54 25 AM

In any event, this implementation does what it says and by my estimate, it's ready to merge with further tweaks coming downstream.

epbrenner (approved):
See [comment](https://github.com/JRaviLab/molevolvr2.0/pull/54#issuecomment-2898840985) for details, but for initial implementation of a scaled phylogenetic tree, I think this hits the mark.

vincerubinetti:
> Eyeballing it, it does seem that the distance addition logic might be a little off.

Currently what it does is just sum the `dist` values for each node in the selected path (shown in green). I'm just using [this d3 function](https://d3js.org/d3-hierarchy/hierarchy#node_path) to find the path, which goes up the tree (if necessary) and back down to get to the target node.

It sounds like what you might be looking for is this horizontal distance:

Screenshot 2025-05-21 at 2 48 09 PM

In which case, this is very easy to calculate because it'd just be the difference between each node's `rootDist` value (its distance from the root node), which I'm already calculating.

> but for the time being, I think only having tip nodes is sufficient, and any internal nodes just won't have the circle shapes designating a selectable node.

To be able to show each intermediate/non-tip node's `dist` value (and full label and etc) in a tooltip, each node needs to have a big enough hover target for accessibility reasons, so we need to keep the circles. They don't necessarily need to be _selectable_, though if it doesn't harm anything, I'd rather just leave the selectability in so I don't have to make some special indication to the user that only tip nodes are selectable but any node is hoverable. Plus it still seems like some user might want to measure between two non-tip nodes.

>  Most of those nodes have zero difference between them, so they sit on that vertical plank side-by-side. I'm sure your tree can handle this, but I just didn't see it occur in the randomization attempts on refreshing the page, so keep an eye out for that.

It should handle this automatically, yes. it would just appear to be a long plank. I've seen this happen with a few nodes, though not a bunch of adjacent ones at the same time.

PR #55, opened 2025-06-05
Upset plot, svg/charting overhaul

vincerubinetti:
Unfortunately this became a big PR because everything here was intertwined, and I wanted to (hopefully) solve all these related issues permanently. So, the diff is very big and difficult to review, and the individual commits won't be very clean or helpful. Instead I will mark certain parts of the code to look at. And know that converting the old SVGs to use the new chart wrapper component involved a lot of copy-pasting large blocks of code. Most importantly, please re-test every visualization component on the Netlify preview.

At a high level, here are the changes:

- implement upset plot
- make generic chart wrapper component that handles most things automatically. auto-fits view box to content. SVG units match DOM units 1:1. see discussion below for how i landed on this particular design.
- abstract out download button with various download formats as its own component. integrate into chart component.
- chart text can be truncated automatically by actual rendered width (not number of chars)
- include chart titles in every chart
- include analysis id (or anything else we want) as part of every filename download. add fake analysis id to testbed page.
- turn the legend component into fully SVG instead of DOM elements. implement simple wrapping strategy.
- make MSA and IPR charts fully SVG, with no DOM elements
- MSA now auto-wraps by default, and accurately (instead of loose estimates). when printing, 
- make charts resizable
- add SVG group elements where appropriate for easier editing in vector software
- consistently put settings vars and Props type at the top of each component file for quick reference
- util func simplifications and refactors

---

If you poke through the commit history, you can see some of the many complicated approaches I experimented with for handling sizing and positioning in the SVGs. 

I wanted to scale the SVG shapes up/down based on the available width but keep the font size matching the document's. This was already sort of implemented before this PR, where the viz components made repeated use of a useSvgTransform hook. But it was verbose, so I first attempted to extract it out to a single hook, and then a component. It required iteratively content-fitting and font-sizing back and forth, which would (usually) converge to a stable configuration. But it was very difficult to find a foolproof way to prevent infinite re-renders or weird edge cases. I also tried leveraging simply defining font size in the root of the SVG that matched the document's, then specifying positions/sizes in terms of `em` units, which the browser could use synchronously with no JS. But there are some limitations to this, like not being able to use relative units in things like `transform`s, and some weird exceptions that caused runaway calculations like when an `em`-sized element exceeded the available width of the container. Trying to implement automatic text truncation made this even more complicated.

The MSA & IPR viz's also required a "fill remaining space" feature, i.e. have a column of labels on the left of a fixed width, and use all the remaining space on the right for the tracks/sequences. Another small problem to solve was that `getBBox`, the method used for fitting the view box to the SVG contents, doesn't work with `clip-path`'ed elements, which is needed for IPR's zoom area.

There's much more that I tried or thought about and didn't commit, which has already left my brain, but I think it's not that valuable.

What I landed on keeps things much simpler. The SVG units always match the DOM units, no exception (hopefully). If there isn't enough room on the page to show the SVG's content, instead of trying to shrink dynamically, it will just overflow and show scrollbars. Not only does this avoid needing iterative fitting, it's probably better UX wise since graphical elements stay in the same position and size -- instead of shrinking to an unreadable size -- and you just need to scroll over to see it all. Also this makes the testbed page much more performant. The "fill remaining" feature is accomplished by providing the available width to the consuming component, and letting it use it however it wants.

---

Regarding the MSA auto-wrapping, currently the code is basically:

```ts
setWrap(40);
print()
setWrap(oldWrap);
```

This is just a loose, hard-coded wrapping point that would generally not overflow the page width when printing portrait mode, US letter.

This PR doesn't change our inability to look inside the print dialog, as discussed in #50 and [this comment](https://github.com/JRaviLab/molevolvr2.0/pull/52#issuecomment-2874220380). But I've taken out the hard-coded `40`. I figure now with the resize handle available on every chart, the user can pre-size the chart to whatever they want before opening the print dialog.

I wanted to mention this because, if you look at the preview on a normal desktop or laptop screen width and have your print dialog on defaults, you'll usually see the MSA overflow the page, and I'm aware of it.


frontend/src/components/MSA.tsx
```diff
@@ -1,247 +1,292 @@
...
+      ]}
+    >
+      {({ width }) => {
+        /** max num of chars that can fit in width */
+        const rowChars = wrap
+          ? Math.max(Math.floor((width - labelWidth) / charWidth), minChars)
+          : length;
+
+        /** split sequence into multiple panels */
+        const panels: [number, number][] = pairs(
```
vincerubinetti, line 99: This code is nearly all the same, though it doesn't look like it from the diff. I previously had a separate "chunk" sub-component, but it is now just all in this render callback. The "chunks" are now called "panels", since I don't know what else to call them ("rows" already refers to each row of the viz).

frontend/src/components/Tree.tsx
```diff
@@ -0,0 +1,291 @@
...
+  dist?: number;
+  /** children items */
+  children?: Item[];
+};
+
+/** link line generator */
+const link = line().curve(curveStepBefore);
+
+/** tree/hierarchy plot */
+const Tree = ({ title, filename = [], data }: Props) => {
```
vincerubinetti, line 46: This is not a new file, it was added in #54, so much of this diff will go away once the tree PR is merged and I rebase this PR.

frontend/src/components/Upset.tsx
```diff
@@ -0,0 +1,272 @@
...
+      label?: string;
+      value: number;
+    }[];
+  };
+  /** cell values */
+  data: (boolean | undefined)[][];
+};
+
+/** upset plot */
+const Upset = ({ title, filename = [], x, y, data }: Props) => {
```
vincerubinetti, line 59: New component to take a look at.

frontend/src/util/dom.ts
```diff
@@ -3,6 +3,44 @@ import { deepMap, onlyText } from "react-children-utilities";
 import { debounce } from "lodash";
 import { sleep } from "@/util/misc";
 
+/** get document root styles */
+const getRootStyles = () => window.getComputedStyle(document.documentElement);
```
vincerubinetti, line 7: Various refactoring of these util funcs to be more modular and sensical. Delete ones that are no longer needed.

falquaddoomi:
Hey @vincerubinetti, nice work on this; it's a pretty ambitious refactor and IMHO makes common functionality a lot more streamlined for these components.

I did a little bit of testing both on the Netlify preview and running the frontend locally. All the testes were on my M1 Macbook running OS X 15.5 ("Sequoia") with the following browsers:
- Chrome (136.0.7103.114, arm64)
- Firefox (139.0.4, aarch64)

I ran into a few issues during testing, mostly with saving charts as PDFs, specifically:
- Saving a chart as PDF resets the scroll position to the top of the page
- When saving anything as a PDF and if the the MSA and/or the IPR charts are in view, their content, but not their frame, will become very narrow when the elements on the page are restored.
- I'm having a hard time reliably reproducing it, but occasionally saving as a PDF will capture the entire page, not just the selected chart. Once it happens once, it happens again for every chart I attempt to save until I refresh the page. This will cause the MSA and IPR charts to become narrow, too, as described before.
- Firefox only: the PNG and JPEG save options don't appear to do anything.

falquaddoomi (commented):
The code looks much improved from before; I can see that centralizing the charting functionality in the Chart component reduces a lot of duplicated effort across the rest of the charting components, so kudos for that.

I made a comment on the PR about some PDF issues I ran into during testing. I'm wondering if they're side effects of the DOM being changed due to the swapping to a portal containing just the element for printing (that I had just said were a good idea...)

If you need additional evidence of the bugs, I can make screen recordings and post them to the PR; just let me know.

vincerubinetti:
Thanks for the in-depth testing and review. I hope that this is the last large scale refactor I have to do, and the experimentation I did makes me confident that this is a good way to construct custom general-purpose SVG charts in the future.

---

>  Also, would it be possible to seed the RNG via a querystring parameter?

Added.

> Regarding the scrollbar: I guess the stakeholders have the important opinion here, but IMO the scrollbar gives you an idea of how zoomed-in you are relative to the full sequence, which I think is important for sequence browsers.

Re-added.

> * Saving a chart as PDF resets the scroll position to the top of the page

Fixed.

> * When saving anything as a PDF and if the the MSA and/or the IPR charts are in view, their content, but not their frame, will become very narrow when the elements on the page are restored.

Fixed. It had something to do with the fact that I was passing the `chart` sub-component, which had the critical `containerRef` attached to it, as a prop to the `Download` component which would then temporarily mount/unmount it as a portal, which for some reason would cause the ref to become `null` and mess up dimensions and everything.

> * I'm having a hard time reliably reproducing it, but occasionally saving as a PDF will capture the entire page, not just the selected chart. Once it happens once, it happens again for every chart I attempt to save until I refresh the page. This will cause the MSA and IPR charts to become narrow, too, as described before.

I couldn't replicate this, but I have a feeling it might have been fixed by the fix above.

> * Firefox only: the PNG and JPEG save options don't appear to do anything.

Looks like this is occurring in the current `main` branch as well, and is likely due to this: https://github.com/bubkoo/html-to-image/issues/508

Unfortunately, setting `skipFonts` just causes the image to be static noise in FF for me. So patching the package might be the only way. Or maybe going back to [dom-to-image-more](https://www.npmjs.com/package/dom-to-image-more) would fix this.

falquaddoomi (approved):

PR #59, opened 2025-06-12
Table of contents and about page rework

vincerubinetti:
- simplify flex component props, e.g. `direction="column"` -> `column`, since it's used everywhere. find+replace whole app.
- change heading component to register/unregister itself in global list of headings
- consistent use of data attribute for active state styles
- tweak table of contents styles
- instead of requerying for all headings on any document mutation (expensive\*), render table of contents from global reactive list of headings
- include icons in table of contents entries
- clean up and compactify about page (original impetus for making TOC look and behave better)

\* On any modern machine though, probably not a humanly perceptible difference. Also the expense will be limited by the fact that we should never have as many things on one page as is currently on the testbed page.

### DIFF
--- a/frontend/src/components/Chart.tsx
+++ b/frontend/src/components/Chart.tsx
@@ -211,7 +211,7 @@ const Chart = ({
   if (printing) return createPortal(chart, document.body);
 
   return (
-    <Flex direction="column" gap="lg" full>
+    <Flex column gap="lg" full>
       {chart}
 
       {/* controls */}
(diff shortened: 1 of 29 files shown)


falquaddoomi (approved):
I gave the code a quick look; the code changes make sense as far as I can tell, and seem cleaner.

IMO the about page looks way more organized than before. I like the new TOC styling, and especially the icons in it for each header. Nice work!

PR #61, opened 2025-06-25
Feedback form

### LINKED ISSUE #56: Feedback form
Something I could add pretty easily, since it's independent of everything else mostly, is a feedback form.

This would be a floating button on the bottom right of every page that opens a feedback form modal, which allows the user to create a GitHub issue right from the webpage, optionally providing their name/email/username, and automatically including info like current page/analysis id/browser/OS/device/etc.

We implemented this for monarchinitiative.org to great utility.

vincerubinetti:
Closes #56 

- pull out repeated static text snippets -- like lab name, email, github org, etc -- into `.env` file variables
- add type defs for env vars
- add random delay to API mock handlers to test loading spinners and such
- move tanstack-query QueryProvider up in app tree so feedback form can use it
- add placeholder api func for submitting feedback form
- tweak dialog component props and styles, including on-open-state-change callback
- add feedback form component in corner of every page
    - optional name, username, email fields
    - sync fields with local storage, reset on successful form submission
    - auto-include browser/OS/device/etc details
    - in case of error, fallback to `mailto:` link with pre-filled subject and body
- use new .env vars wherever appropriate, such as in footer links
- fix link component bug when navigating to root page
- tweak alert component types and styles
- use destructured `import.meta.env` wherever appropriate
- in textbox component, only show copy button on multiline inputs
- make "view corner" component just buttons. move toasts to separate element.
- make note on about page to encourage use of feedback form instead of directly emailing

@falquaddoomi I was initially thinking we'd just do a one-off serverless/cloud function for this. But maybe since we'll have a continuously running server anyway, we can just make this part of the regular backend. But on the other hand maybe it'd be good to have the feedback form as an isolated thing so if the backend goes down for some reason the feedback form still works? If you want to do it as a cloud function, I have a snippet of Node code ready to go (from another project) that I could set up ahead of time. I'll leave it up to you.


vincerubinetti:
> * Why is there a copy button on the text fields? I'm not against it, but I'm not sure I understand the purpose.

I think I had just added it to the textbox component as always there, but it's not usually that useful. Cases where it might be more useful are longer bits of text that the user has typed or pasted and wants to save or move somewhere else for some reason. Perhaps I'll hide it by default for single line textboxes and show it by default for multiline.

> * The screenshot feature is cool, and I understand that you explain that they'll attach it to the resulting discussion on GitHub in the tooltip. Frankly, I missed the tooltip the first time I interacted with it and then found myself wondering how to use the screenshot. I wonder if there's a way to somehow attach it to the issue automatically on creation, but if not, no worries.

Unfortunately GitHub doesn't seem to have an API for uploading images to their `github.com/user-attachments` (images in issues) feature.

https://stackoverflow.com/questions/28703544/create-an-issue-with-image-via-github-api
https://github.com/octokit/octokit.rb/issues/986
https://github.com/isaacs/github/issues/1133
https://stackoverflow.com/questions/76444295/what-happened-to-the-github-content-attachments-api

I'd imagine this is to prevent abuse, or maybe it's just a low priority. So we'd have to do it in a more round-about way: saving it to our own bucket that we maintain in perpetuity, saving it to a third party service like imgur, committing it to one of our repos, etc.

For now I figured we'd just instruct them to do it manually, if needed. I'll tweak the form to make it more clear why that button is there.

PR #62, opened 2025-07-10
Add source for feedback cloud func

vincerubinetti:
Followup to #56 and #61. Adds code for a google cloud function that creates an issue on a particular repo.

Currently this is setup under my own personal GCP project, and hooked up to a test repo under my name (which I'll delete later). You can test by opening the webapp, adding `?mock=false` to the url (to make the app make real network requests), submitting the feedback form, and (hopefully) watching it appear at https://github.com/vincerubinetti/test/issues

Before merge, we should re-follow the setup instructions I've included but with university/Ravilab-associated tokens/repos/gcp projects/etc.

- move the accessing of the load-time url (from address bar or session storage 404 redirect) to one spot in /util/url file
- use /util/url in router redirection
- add frontend api func that hits live cloud func
- reword "discussion/issue" in UI to more vague "post"
- allow request mocking to be controlled from url param, using /util/url
- use /util/url in router redirection in seed
- add setup instructions for cloud func
- add cloud func javascript node code that creates an issue in a particular repo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /.env
 .DS_Store
+node_modules
--- a/frontend/bun.lock
+++ b/frontend/bun.lock
@@ -41,6 +41,7 @@
         "@axe-core/playwright": "^4.10.2",
         "@eslint/js": "^9.30.1",
         "@ianvs/prettier-plugin-sort-imports": "^4.4.2",
+        "@octokit/types": "^14.1.0",
         "@playwright/test": "^1.53.2",
         "@types/cytoscape": "^3.21.9",
         "@types/cytoscape-avsdf": "^1.0.3",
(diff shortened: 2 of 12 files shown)


falquaddoomi (approved):
I was able to submit feedback and see it show up in the test repo's issue tracker with all the information from the submission page. The implementation is clear and the documentation is sufficient to deploy it. Nice work!

vincerubinetti:
> This definitely shouldn't block this PR, but something that occurred to me afterward is that these issues will be created as the user account for which you created the access token. 

Indeed, it's not ideal. I'm fine leaving it as my PAT for now, but eventually I think we should do either an app or a PAT under Janani's name.

I would like to move from my personal GCP project to the cuhealth-ai one, or ideally (later) a RaviLab-controlled GCP, before merging though.

PR #69, opened 2026-01-23
Misc fixes

### LINKED ISSUE #67: Simplify components
Though I've extensively tested all the components, I think there are too many modes of operation and too many points of failure. I tried to make them all support controlled vs uncontrolled, form validation vs non, form data vs declarative data, etc. Maybe I was trying to make this suite of components usable in any type of project, but that was unwise. Instead, I think things can just be simplified by enforcing a particular mode, e.g. controlled. Investigate.

vincerubinetti:
Closes #67 

Apologies, this became a big bucket of random fixes and improvements. But the diff looks worse than it is, due to tailwind re-linting, updating icons, indentation changes (un/wrapping elements), etc. You can safely glaze over those things. Just spot checking the app for equivalent (or better) behavior from main would be good.

This PR also incorporates the lessons learned in https://github.com/krishnanlab/meta2onto/pull/5.

Roughly in order of diff, more important things bolded:

- simplify playwright installation. just install same browsers in whatever context. can turn back on firefox/safari testing when there are more e2e tests.
- **clean up prettier config**, re-add `prettier-plugin-tailwindcss` just to format `.css` files. (more than just formatting, can catch CSS bugs like applying two conflicting classes)
- remove `prettier-plugin-css-order` plugin, as it's barely necessary when 99% of styles are using tailwind
- remove react-refresh eslint plugin. only rule was turned off anyway because warning was annoying, and rule generally isn't that helpful.
- **clean up eslint config** to match current recommended style from each plugin library
- clean up package json scripts
- upgrade package versions, notably: **new restraints from react compiler**, **eslint better tailwind v4** which changes lots of tailwind formatting.
- **use knip to remove some unused files, functions, exports, etc**
- simplify "scroll to hash" logic on route change
- **convert font awesome icons to lucide icons** wherever possible, look nicer and more consistent
- remove tedious use of `type="button"` and instead prevent unwanted behavior more simply in form component
- **address new eslint warnings and constraints due to new react compiler**
- use button component in place of native button element where appropriate
- **remove component local copy of state** and **make components controlled** where possible. reduces complexity and bug potential from having to support controlled/uncontrolled.
- **remove component "layout" prop**. several input components previously had a vertical/horizontal layout prop. now, instead, layout is more appropriately left to the in-situ usage. adds a bit of repetition, but with tailwind being so terse it's not bad. it also allows for more flexibility like putting multiple controls in a grid layout where their labels are aligned.
- **various styling changes** for components and all over the place really
- **get rid of FormData support**. instead, since all components should be controlled, form state should be saved in useState variables outside the form, and passed directly to an API call or etc. simplifies a lot of complexities, e.g. compiling all the values into FormData in an expected and convenient format, and needing extra fake/invisible inputs on components to support it.
- rework table component. turn off react compiler since tanstack library doesn't support it yet. change heading button behavior/appearance a bit.
- rework textbox component to handle side buttons better
- combine mark and toast component types (info/warning/info/etc.)
- use table component for about page team table, and put team info in separate json file
- **rework new analysis page** in accordance with changes to styles and components. e.g., layout blast param controls in vertical grid where labels and inputs are aligned
- **update testbed page** in accordance with changes to styles and components. notable, have state be controlled with useState in each section.
- improve home page bg viz. make more prominent, **fix memory leak**, make more aesthetic
- **reorganize and simplify dom util funcs** for scrolling, css, etc.
- update tsconfig


falquaddoomi (approved):
I skimmed the code, but it all looks reasonable to me.

I like the Lucide icons; I think they fit your aesthetic better than the Font Awesome ones. Hoisting the state out of the form controls is IMO also a good  move.

